PR #1, opened 2019-06-25
feature nex-92 initial copy of qti test runner sources

btamas:
Jira: https://oat-sa.atlassian.net/browse/NEX-92

Test:
`npm run test`

Depend on:
- [x] https://github.com/oat-sa/rollup-plugin-wildcard-external/pull/2
- [x] After merge of rollup-plugin-wildcard PR, `package.json` should be modified

Task after:
- [ ] release it to npm


scss/new-test-runner.scss
```diff
@@ -0,0 +1,17 @@
+@import 'inc/bootstrap';
+@import 'inc/colors';
+// @import "inc/loading-bar";
+// @import "inc/action-bars";
+// @import 'inc/section-container';
```
krampstudio: they are into TAO right ? we need to sort this out

btamas: Yes, all 3 are in `tao-core` and I do not know, how should I load them.

PR #6, opened 2019-07-18
[TAO-8713] NFERlightweight: Passing a test with an offline test runner leads to different and incorrect displaying of the final test score

AntonTsymuk:
Related to : https://oat-sa.atlassian.net/browse/TAO-8713
 
While passing the test with offline test runner without internet connection, if test taker will download test data then it will cause reset of test taker action timestamp

#### How to test
- Run delivery as a test taker with offline test runner
- Break the connection
- At the end of the test download the test data
- Restore the connection and finish the test
- Check that data with correct timestamp sent to /taoQtiTest/OfflineRunner/messages endpoint

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-test-runner-qti",
-    "version": "0.2.1",
+    "version": "0.2.2",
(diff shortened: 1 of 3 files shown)


handleman:
code fine, works well

PR #9, opened 2019-07-29
key navigation shows full content

zagovorichev:
https://oat-sa.atlassian.net/browse/TAO-8693

User can view full response when using tab and arrow keys for navigation.


zagovorichev:
@AntonTsymuk PR fixed, could you check it again?

PR #11, opened 2019-07-31
update tao-core-libs

btamas:
Update tao-core-libs to the new version and remove local libs.

Test:
`npm run test`

Requirement:
- [x] https://github.com/oat-sa/tao-core-libs-fe/pull/6
- [x] update lib version in package.json
- [x] update package-lock.json


btamas:
The test fail was there before. If I will have time, I will check it.

krampstudio:
The PR itself is ok for me, I still need to test it with tao curgen

btamas:
Tests are broken, but Pavel will fix it in his PR.

PR #13, opened 2019-07-31
Bug/tao 8691/line reader resize pointers

handleman:
**task:**  https://oat-sa.atlassian.net/browse/TAO-8691

**description:**
resize controls had wrong cursors, because inner mask resize control was in fact two different controls: vertical resize and horizontal.

**solution:**
one of the controls was disabled and other was extended to be able resize mask both horizontally and vertically.

### DIFF
--- a/scss/inc/_line-reader.scss
+++ b/scss/inc/_line-reader.scss
@@ -50,6 +50,7 @@ $lrMaskBgTransparent: rgba($lrMaskBg, 0.8);
             position: absolute;
             right: 0;
             bottom: 0;
+            cursor: nwse-resize;
         }
     }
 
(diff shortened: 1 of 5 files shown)


scss/inc/_line-reader.scss
```diff
@@ -50,16 +50,20 @@ $lrMaskBgTransparent: rgba($lrMaskBg, 0.8);
...
         .resize-control {
             position: absolute;
-            width: 10px;
+            width: 20px;
             height: 20px;
             bottom: -10px;
+            left: -10px;
             border-right: 1px solid $pluginLink;
+            border-bottom: 1px solid $pluginLink;
+            cursor: nesw-resize;
```
AntonTsymuk: should be `nwse-resize`

PR #18, opened 2019-08-08
Pause if item render error is presented

tikhanovichA:
https://oat-sa.atlassian.net/browse/SI-351
We caught a case when during initialization a page, not all js plugins loaded. Because of that, we had an error on client side wich blocked all actions.
For fixing that case we call pause action if the initialization process has errors.
For that, we added a new property `unrecoverable` for the Error object which marks that we need to set a test to the pause state.

This PR linked with https://github.com/oat-sa/tao-item-runner-qti-fe/pull/11


src/provider/qti.js
```diff
@@ -614,6 +614,9 @@ var qtiProvider = {
                     self.trigger('enablenav');
                     reject(err);
                 })
+                .on('itemrender.error', function (err) {
```
krampstudio: the other war around, please use an namespace on `error` : 
```
on('error.itemrender`, function(err){
```

tikhanovichA: Hello. I just checked and if I will trigger `error.itemrender` then that event will be caught by 
`.on('error.itemrender'` and by `.on('error'`.

krampstudio: I'd suggest then to add the property `unrecoverable` to the `Error`  : 
```
const err = new Error('Item rendering issue');
err.unrecoverable = true;
runner.trigger('error', err);
```

then inside the test runner implementation : 
```
.on('error', err => {
  if(err && err.unrecoverable){
     runner.pause();
  } else {
      //...
  }
})
```

we already are using this formalism for cancellable errors

tikhanovichA: got it. Thanks a lot @krampstudio

src/provider/qti.js
```diff
@@ -666,8 +666,13 @@ var qtiProvider = {
                 assetManager: assetManager
             })
                 .on('error', function(err) {
-                    self.trigger('enablenav');
-                    reject(err);
+                    if(err && err.unrecoverable){
+                        self.trigger('pause', {message : err.message});
+                    } else {
+                        self.trigger('enablenav');
+                        reject(err.message);
```
krampstudio: please reject only the `error` : `reject(err)`

src/provider/qti.js
```diff
@@ -666,8 +666,13 @@ var qtiProvider = {
...
+                    }
+
```
krampstudio: please remove unecessary blank line

PR #25, opened 2019-08-28
Fix/optimize map helper

btamas:
Test: `npm run test`

### DIFF
--- a/src/helpers/map.js
+++ b/src/helpers/map.js
@@ -84,15 +84,21 @@ export default {
      * @returns {Object} the active item
      */
     getActiveItem: function getActiveItem(map) {
-        const parts = this.getParts(map) || {};
+        const parts = this.getParts(map);
 
-        for (let part of Object.values(parts)) {
-            const sections = part.sections || {};
-            for (let section of Object.values(sections)) {
-                const items = section.items || {};
-                for (let item of Object.values(items)) {
-                    if (item.active) {
-                        return item;
+        if (parts) {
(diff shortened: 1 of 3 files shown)


src/helpers/map.js
```diff
@@ -73,18 +73,9 @@ export default {
...
-            if (sections) {
-                _.forEach(sections, function(section) {
-                    result[section.id] = section;
-                });
-            }
-        });
-        return result;
+        const parts = this.getParts(map) || {};
+
+        return Object.values(parts).reduce((result, part) => Object.assign(result, part.sections), {});
```
ferenckiss89: can it happen that we don't have sections?

src/helpers/map.js
```diff
@@ -93,25 +84,21 @@ export default {
...
-                        _.forEach(items, function(item) {
-                            if (item.active) {
-                                result = item;
-                            }
-                        });
+        const parts = this.getParts(map) || {};
+
+        for (let part of Object.values(parts)) {
+            const sections = part.sections || {};
+            for (let section of Object.values(sections)) {
```
ferenckiss89: does it make sense to continue if `sections` is an empty object?

src/helpers/map.js
btamas: Yes it can

src/helpers/map.js
btamas: It won't continue, because empty object has no values. But I will add a condition instead.

PR #33, opened 2019-09-04
[TAO-8934] WTP-1401 - TCiaB : Magnifying glass does not magnify contents correctly.

AntonTsymuk:
Related to : https://oat-sa.atlassian.net/browse/TAO-8934
 
Capture scrolling changes constantly, but not only when magnifier is active
  
#### How to test

- prepare a test item which does not fit browser screen size
- prepare a test with the test item and magnifying test taker tool enabled
- prepare a delivery for the test and execute it as a test taker
- go to the test item and scroll down to the bottom of the page
- activate the magnifying tool and make sure it magnifies correct area of the page

### DIFF
--- a/src/plugins/tools/magnifier/magnifierPanel.js
+++ b/src/plugins/tools/magnifier/magnifierPanel.js
@@ -44,6 +44,16 @@ var debounceDelay = 50;
 var scrollingDelay = 20;
 
 /**
+ * Information about scrooling offset in dom nodes
+ */
+let scrolling = [];
+
+/**
+ * Component scroll handler
+ */
+let componentScrollCallback;
+
+/**
  * The default base size
  * @type {Number}
  */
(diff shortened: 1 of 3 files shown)


src/plugins/tools/magnifier/magnifierPanel.js
```diff
@@ -93,6 +103,39 @@ var dynamicComponentDefaultConfig = {
...
+            id: scrollId,
+            scrollTop: scrollingTop,
+            scrollLeft: scrollLeft
+        });
+    }
+
+    componentScrollCallback && componentScrollCallback(scrollData);
+}, scrollingDelay);
+
+window.addEventListener('scroll', scrollingListenerCallback, true);
```
bziondik: @jsconan  I am not sure about listening scrolls every time even user don't use the plugin.
Maybe there is some way to calculate scroll right before the plugin will show?

jsconan: Indeed, you are right. Even with the throttle applied on the callback, the event should only be applied when needed.

AntonTsymuk: yes, but i did not find another option how to restore scroll position when plugin is activated

jsconan: The issue is the scroll position is not accurate when the plugin is activated. Then why not properly initialize the callback context, without moving it?

Something like:
```
...
    let scrollPosition = getCurrentScrollPosition();
    var scrollingListenerCallback = _.throttle(function(event) {
        var $target = $(event.target);
        var scrollingTop = event.target.scrollTop + scrollPosition.top;
        var scrollLeft = event.target.scrollLeft + scrollPosition.left;
        ...
    }, scrollingDelay);
...
    /**
     * Initializes the listener for scrolling event and transfer the scrolling
     */
    function setScrollingListener() {
        scrollPosition = getCurrentScrollPosition();
        window.addEventListener('scroll', scrollingListenerCallback, true);
    }

    /**
     * Stops the listener for scrolling event
     */
    function removeScrollingListener() {
        window.removeEventListener('scroll', scrollingListenerCallback, true);
    }
...
```

PR #38, opened 2019-09-09
Fix/nex 240 hasfeedbacks also from testmap

krampstudio:
deprecated but backward compatible `hasFeedbaks` property to be declared preferably on the testMap's current item instead of the test context.

How to test : 
- `npm run test`
- run a test with a modal feedback


src/provider/qti.js
```diff
@@ -244,7 +244,9 @@ var qtiProvider = {
          * @param {Promise} [loadPromise] - wait this Promise to resolve before loading the item.
          */
         function computeNext(action, params, loadPromise) {
-            var context = self.getTestContext();
+            const context = self.getTestContext();
+            const testMap = self.getTestMap();
```
btamas: `testMap` is defined but never used.

PR #41, opened 2019-09-11
Fix/nex 240 deprecate qti item info in context

krampstudio:
- deprecate the usage `allowComment`, `allowSkipping` and `validateResponse` from the testContext. Take them from the `testMap`'s current item instead.

How to test : 
 - `npm run test`
 - Try a regressions on a test with the configured properties `allowComment`, `allowSkipping` and `validateResponse` on some items


krampstudio:
good catch I've removed them

PR #54, opened 2019-09-30
Fix/nfersup 25/pause on unrecoverable error

hutnikau:
Related to : https://github.com/oat-sa/extension-tao-testqti/pull/1606
Please see the descriptions in #1606

### DIFF
--- a/src/proxy/cache/proxy.js
+++ b/src/proxy/cache/proxy.js
@@ -212,9 +212,9 @@ export default _.defaults(
                                                         : null;
 
                                                 if (!actionResult.success && self.actionRejectPromises[actionId]) {
-                                                    var error = new Error(actionResult.message);
+                                                    const error = new Error(actionResult.message);
                                                     error.unrecoverable = true;
-                                                    reject(error);
+                                                    return reject(error);
                                                 }
 
                                                 if (actionId && self.actionPromises[actionId]) {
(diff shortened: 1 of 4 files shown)


src/proxy/cache/proxy.js
```diff
@@ -207,6 +211,12 @@ export default _.defaults(
                                                         ? actionResult.requestParameters.actionId
                                                         : null;
 
+                                                if (!actionResult.success && self.actionRejectPromises[actionId]) {
+                                                    var error = new Error(actionResult.message);
```
krampstudio: `const error`

src/proxy/cache/proxy.js
```diff
@@ -207,6 +211,12 @@ export default _.defaults(
...
+                                                    error.unrecoverable = true;
+                                                    reject(error);
```
krampstudio: please return before `reject` to prevent unexpected code execution

zagovorichev:
Not sure that I did everything properly, but:
1. Started new delivery execution

2. open second tab with the same de
3. on the 1 tab I answered yes no yes and press next:

4. on the second tab answered no yes no for the same first item:

nothing was changed
5. 2 tab: go back and do the same no yes no:

zagovorichev:
@hutnikau check my comments, please

PR #56, opened 2019-10-04
[TAO-8934] WK-1401 - TCiaB : Magnifying glass does not magnify contents correctly.

AntonTsymuk:
Related to : https://oat-sa.atlassian.net/browse/TAO-8934
 
Full description of changes.
  
#### How to test
 
- prepare instance of TAO with taoAct extension 
- prepare a delivery with enabled magnifying glass tool
- run delivery and check that there is no transparent space at the bottom
- change debug mode to false and check one again

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-test-runner-qti",
-    "version": "0.10.6",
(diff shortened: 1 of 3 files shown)


handleman:
code fine, works well

PR #67, opened 2019-10-30
sending message to the server in English language

AndreyNegrov:
message was sent to the server in the language of the test taker
now the message is sent to the server in English language


AndreyNegrov:
> no version bump

fixed

PR #68, opened 2019-11-05
[TAO-9473] WTP-1466 - [ACT] [TTS - FE] Create TTS plugin

AntonTsymuk:
Related to : https://oat-sa.atlassian.net/browse/TAO-9474, https://oat-sa.atlassian.net/browse/TAO-9476
 
Create adapter for new APIP Text To Speech plugin. Update keyNavigation plugin to be able to handle TTS plugin tab navigation

#### How to test
 
- Go to `config/taoTests/test_runner_plugin_registry.conf.php` and change `active` to `true` for APIP Text To Speech plugin
- Prepare an instance of tao with taoAct extension
- Prepare a test with enabled APIP Text To Speech plugin
- Execute test as a test taker
- Make sure that APIP Text To Speech plugin is available in custom toolbox

Requires :
 - [ ] https://github.com/oat-sa/extension-tao-act/pull/1258
 - [ ] https://github.com/oat-sa/extension-tao-testqti/pull/1626

### DIFF
--- a/src/plugins/content/accessibility/keyNavigation.js
+++ b/src/plugins/content/accessibility/keyNavigation.js
@@ -526,7 +526,7 @@ function initDefaultItemNavigation(testRunner) {
 function initTTSNavigation(testRunner) {
     const pluginName = 'apiptts';
     const actionPrefix = `tool-${pluginName}-`;
-    const $itemElement = $(`[data-control="${pluginName}"]`);
+    const $itemElement = $(`.${pluginName}-plugin`);
     const id = `${pluginName}_navigation_group`;
 
     const test1Navigator = keyNavigator({
@@ -693,14 +693,14 @@ export default pluginFactory({
                 pluginConfig.contentNavigatorType = type;
             })
             .on('plugin-open.apiptts', () => {
-                const $itemElement = $('[data-control="apiptts"]');
+                const $itemElement = $('.apiptts-plugin');
 
                 $itemElement.addClass(ignoredClass);
 
                 this.groupNavigator.goto('apiptts_navigation_group');
             })
             .on('plugin-close.apiptts', () => {
-                const $itemElement = $('[data-control="apiptts"]');
+                const $itemElement = $('.apiptts-plugin');
 
                 $itemElement.removeClass(ignoredClass);
             });
--- a/src/plugins/content/accessibility/keyNavigation.js
+++ b/src/plugins/content/accessibility/keyNavigation.js
@@ -47,12 +47,6 @@ var defaultPluginConfig = {
 };
 
 /**
- * initialized test runner navigator
- * @type {Object}
- */
-let testRunnerNavigatorItem;
-
-/**
  * Init the navigation in the toolbar
  *
  * @param {Object} testRunner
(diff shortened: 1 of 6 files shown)


src/plugins/tools/apipTextToSpeech/plugin.js
```diff
@@ -0,0 +1,209 @@
...
+        /**
+         * Show the plugin panel
+         *
+         * @fires plugin-open.apiptts
+         */
+        const enablePlugin = () => {
+            this.button.turnOn();
+            this.setState('active', true);
+
+            testRunner.trigger(`plugin-open.${pluginName}`);
```
jsconan: no need to use this, you may directly use the plugin alias: `this.trigger('open')`

https://github.com/oat-sa/tao-core-sdk-fe/blob/master/src/core/plugin.js#L195

src/plugins/tools/apipTextToSpeech/plugin.js
```diff
@@ -0,0 +1,209 @@
...
+        /**
+         * Hide the plugin panel
+         *
+         * @fires plugin-close.apiptts
+         */
+        const disablePlugin = () => {
+            this.setState('active', false);
+
+            this.button.turnOff();
+            testRunner.trigger(`plugin-close.${pluginName}`);
```
jsconan: Same here: `this.trigger('close')`

src/plugins/tools/apipTextToSpeech/plugin.js
```diff
@@ -0,0 +1,209 @@
...
+/**
+ * Returns the configured plugin
+ */
+export default pluginFactory({
+    name: pluginName,
+
+    /**
+     * Initialize the plugin (called during runner's init)
+     */
+    init: function init() {
```
jsconan: Why not use object shorthand instead?
```javascript
init() {
   ...
},
render() {
    ...
}
```

src/plugins/content/accessibility/keyNavigation.js
```diff
@@ -523,6 +523,30 @@ function initDefaultItemNavigation(testRunner) {
     return itemNavigators;
 }
 
+function initTTSNavigation(testRunner) {
+    const pluginName = 'apiptts';
+    const actionPrefix = `tool-${pluginName}-`;
```
jsconan: Please avoid introducing coupling like that. APIP TTS is a plugin, and therefore should not be directly expected by a generic component.

Eventually, you may add another generic group. But not hard coupled.

As an idea, you might create a group based on a particular css class. And use this class to tag the plugin elements. This is how this is made for any other groups.

src/plugins/content/accessibility/keyNavigation.js
```diff
@@ -665,6 +691,18 @@ export default pluginFactory({
              */
             .on('setcontenttabtype', function(type) {
                 pluginConfig.contentNavigatorType = type;
+            })
+            .on('plugin-open.apiptts', () => {
+                const $itemElement = $('[data-control="apiptts"]');
+
+                $itemElement.addClass(ignoredClass);
+
+                this.groupNavigator.goto('apiptts_navigation_group');
```
jsconan: Same here. Please avoid introducing coupling.

AntonTsymuk:
@jsconan i found another way how we can handle key navigation. Instead of modifying keyNavigation, we can create a navigation group inside TTS plugin. With this way we will not have any dependencies between TTS and keyNavigation plugins. Please take a look

src/plugins/tools/apipTextToSpeech/plugin.js
```diff
@@ -0,0 +1,238 @@
...
+                const $navigationGroupElement = this.button.getElement();
+                const groupNavigationId = `${pluginName}_navigation_group`;
+
+                if (!$navigationGroupElement) {
+                    return;
+                }
+
+                this.navigationGroup = keyNavigator({
+                    id: groupNavigationId,
+                    elements: navigableDomElement.createFromDoms($navigationGroupElement),
```
jsconan: I'm curious about the elements you need to navigate through. Are they the menus or are they the audio elements?
Here it look like they are the menus since they come from the tool button.

AntonTsymuk: User will be able to navigate between content parts with associated media files. From the technical requirements:
the user can use their keyboard to focus on the sequence of access elements (display the hover effect on focus) by tabbing through the content chunks (shift-TAB to go backwards in the sequence). Users may initiate playback from the current focused content chunk by pressing ENTER or SPACEBAR.
The requirements can be found here: https://oat-sa.atlassian.net/wiki/spaces/UX/pages/118522064/TAO+9016+-+SPIKE+-+Story+breakout+for+TTS+Text+to+Speech
Also this will relay on APIP test item data. The APIP data includes the order for the content parts and we will need to navigate between them according to this order. The idea is to listen to tab navigation events and change items dynamically according to the order, but it will be solved during implementation of the component for this plugin

jsconan: ok, got it

PR #72, opened 2019-11-14
Fix/tao 9561/in case offline mode don't throw error

bziondik:
Related to: https://oat-sa.atlassian.net/browse/TAO-9561

In the case with connectivity error don't throw error. This will fix the case when user stays on the same item after losing connection to the server.

### DIFF
--- a/src/proxy/offline/proxy.js
+++ b/src/proxy/offline/proxy.js
@@ -318,7 +318,6 @@ export default _.defaults(
                             }
 
                             self.trigger('error', err);
-
                             throw err;
                         });
                 });
--- a/src/proxy/offline/proxy.js
+++ b/src/proxy/offline/proxy.js
@@ -304,11 +304,13 @@ export default _.defaults(
                         .flush()
                         .then(function(data) {
                             actions = data;
+                            self.syncInProgress = false;
                             if (data && data.length) {
                                 return self.sendSyncData(data);
                             }
                         })
                         .catch(function(err) {
+                            self.syncInProgress = false;
                             if (self.isConnectivityError(err)) {
                                 self.setOffline('communicator');
                                 _.forEach(actions, function(action) {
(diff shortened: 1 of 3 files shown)


src/proxy/offline/proxy.js
```diff
@@ -307,21 +307,17 @@ export default _.defaults(
                             if (data && data.length) {
                                 return self.sendSyncData(data);
                             }
-                        }).finally((data) => {
-                            self.syncInProgress = false;
```
jsconan: are you sure about this removal? the value is never reset once it has been set to true.
Maybe should you replace `finally` by `then` and move it after the `catch`

krampstudio: Yes better adding it in `then` and `catch` to be safe

bziondik: put `self.syncInProgress = false;` both in then and catch

src/proxy/offline/proxy.js
```diff
@@ -304,24 +304,22 @@ export default _.defaults(
                         .flush()
                         .then(function(data) {
                             actions = data;
+                            self.syncInProgress = false;
```
jsconan: this should be move after the promise is resolved, nor before

jsconan: something like:
```javascript
return this.queue.serie(function() {
                    return self.actionStore
                        .flush()
                        .then(function(data) {
                            actions = data;
                            if (data && data.length) {
                                return self.sendSyncData(data);
                            }
                        })
                        .catch(function(err) {
                            if (self.isConnectivityError(err)) {
                                self.setOffline('communicator');
                                _.forEach(actions, function(action) {
                                    self.actionStore.push(action.action, action.parameters, action.timestamp);
                                });
                                return;
                            }

                            self.syncInProgress = false;
                            self.trigger('error', err);

                            throw err;
                        })
                        .then(data => {
                            self.syncInProgress = false;
                            return data;
                        } );
                });
```

bziondik: fixed

PR #75, opened 2019-11-15
Fix/tao 9603/alllow keyboard on offline sync modal

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-9603

Allows the keyboard navigation on the wait dialog displayed at the end of an offline test when the connectivity is down.

Step to check:
- take a test with the offline mode set up
- pass through the test, before the end disconnect from the network
- try to end the test, the wait popup should be diplayed
- wait for the countdown to finish
- try to reach the download button

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-test-runner-qti",
-    "version": "0.10.14",
+    "version": "0.10.15",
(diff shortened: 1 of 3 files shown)


bziondik:
@js Checked again. This case only if you already downloaded json and then reconnected. 
If not press Download all works.

PR #78, opened 2019-11-19
check is online before resolve offlineAction

bziondik:
Related to: https://oat-sa.atlassian.net/browse/TAO-9666

In `offlineSyncMode` was added handler on `proxy.before('disconnect'`
On `offline/proxy `returned promise of `offlineAction` will be resolved only in online mode
@js how to open closed (by click Proceed&end essesment button) dialog again?

Steps to reproduce:
1) start taking a test;
2) move to offline test taking mode;
3) at the end of the test, come back online and see the message "You are encountering a prolonged connectivity loss. The connection seems to be back, please proceed." Don't click on it.
4) go offline and see the message in red. Don't interact with the message, the message disappears. 
5)while still offline, click on Proceed & End Assessment. -> browser returns standard offline (You're not connected) error message.

### DIFF
--- a/src/helpers/offlineSyncModal.js
+++ b/src/helpers/offlineSyncModal.js
@@ -70,9 +70,12 @@ function offlineSyncModalFactory(proxy) {
 
             proxy.after('reconnect', () => waitingDialog.endWait());
             proxy.before('disconnect', () => {
-                // need to open dialog again!!!
-                // waitingDialog.getElement().modal('open'); - not working 
-                waitingDialog.beginWait()
+                // need to open dialog again if it is closed
+                const waitingDialogModal = $('.preview-modal-feedback');
+                if (!waitingDialogModal.hasClass('opened')){
+                    waitingDialogModal('open');
+                }
+                waitingDialog.beginWait();
             });
 
             // if render comes before beginWait:
--- a/src/helpers/offlineSyncModal.js
+++ b/src/helpers/offlineSyncModal.js
@@ -48,6 +48,7 @@ function offlineSyncModalFactory(proxy) {
         width: '600px'
     };
     let $secondaryButton;
+    const betweenButtonTextSelector = 'div.preview-modal-feedback.modal .between-buttons-text';
     const secondaryButtonWait = 60; // seconds to wait until it enables
     let delaySec;
     const $countdown = $(offlineSyncModalCountdownTpl());
@@ -91,8 +92,7 @@ function offlineSyncModalFactory(proxy) {
             dialogShortcut.clear();
         })
         .on('wait', () => {
-
-            hider.show('.between-buttons-text');
+            hider.show(betweenButtonTextSelector);
             // if beginWait comes before render:
             if (waitingDialog.is('rendered')) {
                 waitingDialog.trigger('begincountdown');
@@ -121,7 +121,7 @@ function offlineSyncModalFactory(proxy) {
             countdownPolling.stop();
             $secondaryButton.prop('disabled', true);
             $countdown.html('');
-            hider.hide('.between-buttons-text');
+            hider.hide(betweenButtonTextSelector);
         });
 
     return waitingDialog;
--- a/src/helpers/offlineSyncModal.js
+++ b/src/helpers/offlineSyncModal.js
@@ -48,7 +48,7 @@ function offlineSyncModalFactory(proxy) {
         width: '600px'
     };
     let $secondaryButton;
-    const betweenButtonTextSelector = 'div.preview-modal-feedback.modal .between-buttons-text';
+    const betweenButtonTextSelector = '.between-buttons-text';
     const secondaryButtonWait = 60; // seconds to wait until it enables
     let delaySec;
     const $countdown = $(offlineSyncModalCountdownTpl());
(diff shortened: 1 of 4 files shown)


src/helpers/offlineSyncModal.js
```diff
@@ -63,11 +64,16 @@ function offlineSyncModalFactory(proxy) {
...
+            delaySec = secondaryButtonWait;
             $secondaryButton = $('div.preview-modal-feedback.modal').find('button[data-control="secondary"]');
-
             $countdown.insertAfter($secondaryButton);
 
-            proxy.off('reconnect.waiting').after('reconnect.waiting', () => waitingDialog.endWait());
+            proxy.after('reconnect', () => waitingDialog.endWait());
+            proxy.before('disconnect', () => {
+                // need to open dialog again!!!
+                // waitingDialog.getElement().modal('open'); - not working
```
bziondik: @js here I need to open waitingDialog again (if needed)

src/helpers/offlineSyncModal.js
```diff
@@ -83,15 +92,17 @@ function offlineSyncModalFactory(proxy) {
             dialogShortcut.clear();
         })
         .on('wait', () => {
+
+            hider.show('.between-buttons-text');
```
jsconan: What are you targeting here? what is the scope?
it looks like a global page element

src/proxy/offline/proxy.js
```diff
@@ -178,30 +178,43 @@ export default _.defaults(
...
-                            }).catch(function() {
-                                return resolve({ success: false });
-                            });
+                                });
+                            };
+                        if (isOffline) {
+                            return new Promise(offlineSync)
+                                .catch(function() {
+                                    return resolve({ success: false });
+                                });
```
jsconan: Not sure to get the idea behind that. `offlineSync()` returns nothing, and therefore the promise catch is useless. Or maybe `offlineSync()` should return a promise?

src/proxy/offline/proxy.js
```diff
@@ -178,30 +178,43 @@ export default _.defaults(
...
                             return self
                                 .syncData()
                                 .then(function() {
+                                    if (self.isOffline()) {
+                                        // in case last request was failed and connection lost
+                                        // show offlineWaitingDialog
+                                        return new Promise(offlineSync)
+                                            .catch(function() {
+                                                return resolve({ success: false });
+                                            });
```
jsconan: Same here. What is the purpose of the void promise?
Moreover, the promise is already caught line 220.

src/proxy/offline/proxy.js
```diff
@@ -178,30 +178,43 @@ export default _.defaults(
...
+                                    self.initiateDownload().catch(function() {
+                                        return resolve({ success: false });
                                     });
-                            }).catch(function() {
-                                return resolve({ success: false });
-                            });
+                                });
+                            };
+                        if (isOffline) {
+                            return new Promise(offlineSync)
```
jsconan: This is useless since the `return` is inside a promise constructor. It won't do anything, the promise won't be consumed. `return offlineSync()` will do the same

src/helpers/offlineSyncModal.js
```diff
@@ -78,20 +85,23 @@ function offlineSyncModalFactory(proxy) {
             dialogShortcut.enable();
         })
         .on('destroy', () => {
+            proxy.off('reconnect.waiting');
+            proxy.off('disconnect.waiting');
```
jsconan: You might also do `proxy.off('.waiting');`

src/helpers/offlineSyncModal.js
```diff
@@ -78,20 +85,23 @@ function offlineSyncModalFactory(proxy) {
...
             globalShortcut.enable();
             dialogShortcut.disable();
             dialogShortcut.clear();
         })
         .on('wait', () => {
+            hider.show(betweenButtonTextSelector);
```
jsconan: Still the same issue :)
Should be rather something like: `hider.show(waitingDialog.dialog.getDom().find(betweenButtonTextSelector));`
But a shorter code should be better ;)

src/helpers/offlineSyncModal.js
```diff
@@ -110,8 +120,8 @@ function offlineSyncModalFactory(proxy) {
         .on('unwait', () => {
             countdownPolling.stop();
             $secondaryButton.prop('disabled', true);
-            $countdown.remove();
-            hider.hide('.between-buttons-text');
+            $countdown.html('');
+            hider.hide(betweenButtonTextSelector);
```
jsconan: Same here

src/helpers/offlineSyncModal.js
```diff
@@ -63,11 +65,16 @@ function offlineSyncModalFactory(proxy) {
     //creates the waiting modal dialog
     const waitingDialog = waitingDialogFactory(waitingConfig)
         .on('render', () => {
+            delaySec = secondaryButtonWait;
             $secondaryButton = $('div.preview-modal-feedback.modal').find('button[data-control="secondary"]');
```
jsconan: didn't notice it, but this is also an issue. The selector should target the proper dialog. What will happen if another dialog is open at the same time?

PR #86, opened 2019-11-28
[TAO-9479][TAO-9480] [ACT] [TTS - FE] Create UI part of the component and make it draggable 

AntonTsymuk:
Related to : https://oat-sa.atlassian.net/browse/TAO-9479, https://oat-sa.atlassian.net/browse/TAO-9480

Add Text To Speech UI component and make it draggable

#### How to test
- Prepare an instance of tao with taoAct extension
- Go to `config/taoTests/test_runner_plugin_registry.conf.php` and change `active` to `true` for APIP Text To Speech plugin
- Prepare an instance of tao with taoAct extension
- Prepare a test with enabled APIP Text To Speech plugin
- Execute test as a test taker
- Enable TTS plugin and check the component
- Make sure that it fits design requirements
- Make sure that component is draggable

### DIFF
--- a/src/plugins/tools/apipTextToSpeech/plugin.js
+++ b/src/plugins/tools/apipTextToSpeech/plugin.js
@@ -27,6 +27,7 @@ import pluginFactory from 'taoTests/runner/plugin';
 import mapHelper from 'taoQtiTest/runner/helpers/map';
 import keyNavigator from 'ui/keyNavigation/navigator';
 import navigableDomElement from 'ui/keyNavigation/navigableDomElement';
+import ttsComponentFactory from 'taoQtiTest/runner/plugins/tools/apipTextToSpeech/textToSpeech';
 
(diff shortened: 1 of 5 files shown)


src/plugins/tools/apipTextToSpeech/plugin.js
```diff
@@ -45,6 +46,24 @@ export default pluginFactory({
...
+
+        /**
+         * Creates the tts component on demand
+         * @returns {textToSpeech}
+         */
+        const getTTSComponent = () => {
+            if (!ttsComponent) {
+                const $container = testRunner.getAreaBroker().getContainer();
+
+                ttsComponent = ttsComponentFactory($container.parent(), {})
```
jsconan: Is it required to place the component on the container's parent? It is unsafe, and it should be safer to not place it higher than the container.

PR #89, opened 2019-12-05
fix/TAO-9730/build jumpTable in proxy init

bziondik:
Related to: https://oat-sa.atlassian.net/browse/TAO-9730

Previously ` offlineNavigation.init() -> offlineJumpTableHelper.init() ` were called every navigation action
Now only in `offline/proxy init ` function. Then only `navigate` function that added new jumps in table.

For correct jumpTuble in `buildJumpTable` function were done (**have doubts about these steps**):
1.  `itemResponse` made from `itemState`
2.  `itemDefinition: item.itemIdentifier` for correct work `addResponsesToResponseStore`

### DIFF
--- a/src/proxy/offline/proxy.js
+++ b/src/proxy/offline/proxy.js
@@ -418,11 +418,7 @@ export default _.defaults(
                             .setTestMap(response.testMap)
                             .init()
                     })
-                    .then(function() {
-                        return new Promise(function(resolve) {
-                            resolve(response);
-                        });
-                    });
+                    .then(() => response);
             });
         },
 
--- a/src/proxy/offline/proxy.js
+++ b/src/proxy/offline/proxy.js
@@ -134,6 +134,7 @@ export default _.defaults(
                      * performs navigation trough items of given test parameters according to action parameters
                      * doesent need active internet connection
                      * @param navigator - navigator helper used with this proxy
+                     * @param {Object} options - options to manage the navigation
                      * @param {Object} options.testContext - current test testContext dataset
                      * @param {Object} results - navigtion result output object
                      */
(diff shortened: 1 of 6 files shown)


src/proxy/offline/proxy.js
```diff
@@ -423,11 +411,18 @@ export default _.defaults(
...
+                    .then(() => {
+                        return self.offlineNavigator
+                            .setTestContext(response.testContext)
+                            .setTestMap(response.testMap)
+                            .init()
+                    })
+                    .then(function() {
+                        return new Promise(function(resolve) {
+                            resolve(response);
+                        });
```
jsconan: No need to wrap promises.
Just do:
```javascript
.then(() => response);
```

src/proxy/offline/proxy.js
```diff
@@ -134,17 +134,13 @@ export default _.defaults(
                      * performs navigation trough items of given test parameters according to action parameters
                      * doesent need active internet connection
                      * @param navigator - navigator helper used with this proxy
-                     * @param {Object} options - options to manage the navigation
                      * @param {Object} options.testContext - current test testContext dataset
                      * @param {Object} results - navigtion result output object
                      */
-                    var navigate = function(navigator, options, results) {
```
jsconan: I'm ok with the remival of the `init`. However, the testContext should be set each time, as it is dependent to the current item.

bziondik: fixed

src/proxy/offline/proxy.js
```diff
@@ -134,17 +134,15 @@ export default _.defaults(
                      * performs navigation trough items of given test parameters according to action parameters
                      * doesent need active internet connection
                      * @param navigator - navigator helper used with this proxy
-                     * @param {Object} options - options to manage the navigation
```
jsconan: I guess you should keep that line as well :)

jsconan:
Looks like there is an issue with the CI/CD.
It looks like related to a system issue.
Moreover, the unit tests work locally.
So merging.

PR #93, opened 2019-12-12
Fix/tao 9435/wrong remaining time on proctor scr

augustas:
Necessary changes to send current item duration when assessment is paused. 
 
Related to : https://oat-sa.atlassian.net/browse/TAO-9435

On `pause` event adds the `itemDuration` parameters to the request payload.
When `teststate` change event with type `pause` is detected during regular sync, then `pause` event is triggered.
 
#### Steps to reproduce
 - Launch an assessment.
 - Go to timed section.
 - In the assessment that I used max time is 00:55:00
 - On item 1, let the timer run down for a few mins.
 - I paused the test when remaining time was 00:52:20
 - The proctor screen, click pause, select a category, sub-category and add description. Click on OK.
 - Notice that remaining time on proctor screen is now 00:55:00.
  
#### Scenarios to test
 
1) Configure a delivery that requires proctoring and has `Security plugins` enabled. Launch it with test taker, create out of focus event by f.e. navigating to another browser tab or another application. Refresh the delivery execution list on proctor screen, notice that test session remaining time matches the time on the client side.

2) Configure a delivery that requires proctoring. Launch it with test taker. Navigate to proctor screen and pause the test, wait for the client side to detect the pause. Refresh the delivery execution list on proctor screen, notice that test session remaining time is correctly reduced.
  
#### Dependencies
 
   
Companion PR :
 - [ ] https://github.com/oat-sa/extension-tao-testqti/pull/1644

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-test-runner-qti",
-    "version": "0.12.0",
+    "version": "0.12.1",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-test-runner-qti",
-    "version": "0.12.0",
+    "version": "0.12.1",
     "description": "TAO Test Runner QTI implementation",
     "files": [
         "dist",
--- a/src/plugins/controls/duration/duration.js
+++ b/src/plugins/controls/duration/duration.js
@@ -108,7 +108,7 @@ export default pluginFactory({
                     self.enable();
                 })
 
-                .before('move skip exit timeout', function() {
+                .before('move skip exit timeout pause', function() {
                     var context = testRunner.getTestContext();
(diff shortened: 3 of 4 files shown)


zagovorichev:
@augustas  we have template example here: https://oat-sa.atlassian.net/wiki/spaces/DEV/pages/11338720/Development+Workflow+with+Github
Could you update description, please?

augustas:
> @augustas we have template example here: https://oat-sa.atlassian.net/wiki/spaces/DEV/pages/11338720/Development+Workflow+with+Github
> Could you update description, please?

Thanks for the suggestion, description updated!

zagovorichev:
@augustas you've added the case when tt paused the test with test runner, but what if tt went away from the page, or just reload it with ctrl+R or F5 (win)?

zagovorichev:
after the discussion about the page reloading issue we decided that that is out of the scope of current task. Current fix fixes the issue with update on `pause`, good job, thanks Augustas

zagovorichev:
@augustas could you update the version, please?

augustas:
> @augustas could you update the version, please?

@zagovorichev, thanks for the review, version bump updated

PR #94, opened 2019-12-12
Fix TAO-9730 handle attempt in offline navigator

bziondik:
Related to: https://oat-sa.atlassian.net/browse/TAO-9730

Implemented handling of `textContext.attempt` in `offlineNavigator`
`attempt`now  is stored in `itemStore`

### DIFF
--- a/src/helpers/testContextBuilder.js
+++ b/src/helpers/testContextBuilder.js
@@ -95,7 +95,6 @@ function getTestContext(testContext, testMap, item, section, part, position, att
                 itemIdentifier: item.id,
                 itemPosition: position,
                 remainingAttempts: Math.max(-1, item.remainingAttempts - 1),
-                attempt: attempt ? attempt + 1 : 1,
                 sectionId: section.id,
                 sectionTitle: section.label,
                 testPartId: part.id,
@@ -104,6 +103,10 @@ function getTestContext(testContext, testMap, item, section, part, position, att
             testContext
         );
 
+    if (attempt) {
+        newTestContext.attempt = attempt + 1;
+    }
+    
     if (isLeavingSection) {
         newTestContext.numberRubrics = 0;
         newTestContext.rubrics = '';
--- a/src/helpers/testContextBuilder.js
+++ b/src/helpers/testContextBuilder.js
@@ -31,7 +31,7 @@ import navigationHelper from 'taoQtiTest/runner/helpers/navigation';
  * @param {Integer} position
  * @returns {Object}
  */
-function buildTestContextFromPosition(testContext, testMap, position) {
+function buildTestContextFromPosition(testContext, testMap, position, attempt) {
     var updatedMap = mapHelper.updateItemStats(testMap, position),
         item = mapHelper.getItemAt(updatedMap, position),
         section = mapHelper.getItemSection(updatedMap, position),
@@ -41,7 +41,7 @@ function buildTestContextFromPosition(testContext, testMap, position) {
         return false;
     }
 
-    return getTestContext(testContext, testMap, item, section, part, position);
+    return getTestContext(testContext, testMap, item, section, part, position, attempt);
 }
 
 /**
@@ -56,12 +56,12 @@ function buildTestContextFromPosition(testContext, testMap, position) {
  * @param {Integer} jump.position
  * @returns {Object}
  */
-function buildTestContextFromJump(testContext, testMap, jump) {
+function buildTestContextFromJump(testContext, testMap, jump, attempt) {
     var part = testMap.parts[jump.part],
         section = part.sections[jump.section],
         item = section.items[jump.item];
 
-    return getTestContext(testContext, testMap, item, section, part, jump.position);
+    return getTestContext(testContext, testMap, item, section, part, jump.position, attempt);
 }
 
 /**
@@ -87,7 +87,7 @@ function buildTestContextFromJump(testContext, testMap, jump) {
  * @returns {Object} the new test context
  * @private
  */
-function getTestContext(testContext, testMap, item, section, part, position) {
+function getTestContext(testContext, testMap, item, section, part, position, attempt) {
     var isLeavingSection = section.id !== testContext.sectionId,
         isLeavingPart = part.id !== testContext.testPartId,
         newTestContext = _.defaults(
@@ -95,6 +95,7 @@ function getTestContext(testContext, testMap, item, section, part, position) {
                 itemIdentifier: item.id,
                 itemPosition: position,
                 remainingAttempts: Math.max(-1, item.remainingAttempts - 1),
+                attempt: attempt ? attempt + 1 : 1,
                 sectionId: section.id,
                 sectionTitle: section.label,
                 testPartId: part.id,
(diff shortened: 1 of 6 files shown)


src/navigator/offlineNavigator.js
```diff
@@ -31,7 +31,8 @@ import testContextBuilder from 'taoQtiTest/runner/helpers/testContextBuilder';
 export default function offlineNavigatorFactory(itemStore, responseStore) {
     var testContext,
         testMap,
-        offlineJumpTableHelper = offlineJumpTableFactory(itemStore, responseStore);
+        offlineJumpTableHelper = offlineJumpTableFactory(itemStore, responseStore),
+        itemStore = itemStore;
```
jsconan: What is the purpose of this variable? it is already given as a parameter.
There are two issues:
- the variable is redeclared with the same name
- the new variable is not assigned later, so keeping the former one should be enough

src/navigator/offlineNavigator.js
```diff
@@ -110,8 +111,24 @@ export default function offlineNavigatorFactory(itemStore, responseStore) {
...
+                        // 4. return new textContext with right attempt
+                        Promise.all([
+                            testContextBuilder.buildTestContextFromJump(testContext, testMap, lastJump),
+                            itemStore.get(lastJump.item)
+                        ])
+                            .then(res => {
+                                const newTestContext = res[0];
+                                const itemFromStore = res[1];
+                                const newAttempt = itemFromStore.attempt ? itemFromStore.attempt + 1 : 1;
+                                newTestContext.attempt = newAttempt;
```
jsconan: Shouldn't it be performed by the testContextBuilder instead? The remaining number is already computed here: https://github.com/oat-sa/tao-test-runner-qti-fe/blob/17475a8a8f9ee721d5e23278f3dd9a08331588c9/src/helpers/testContextBuilder.js#L97
So, it will be better to also increase the attempt at the same time.
And then, you might update the item with the attempt value provided by the context

src/navigator/navigator.js
```diff
@@ -49,62 +49,66 @@ var navigatorFactory = function navigatorFactory(testContext, testMap) {
...
+         * @returns {Promise} - the result of the nav
          */
         navigate: function navigate(direction, scope, position) {
-            var methodName = direction.toLowerCase() + scope.substr(0, 1).toUpperCase() + scope.substr(1).toLowerCase();
+            return new Promise(function(resolve, reject) {
+                var methodName = direction.toLowerCase() + scope.substr(0, 1).toUpperCase() + scope.substr(1).toLowerCase();
+
+                if (_.isFunction(this[methodName])) {
+                    this[methodName](position, attempt)
+                        then(newTestContext => resolve(newTestContext))
```
jsconan: issue there: missing `.`

src/navigator/navigator.js
```diff
@@ -49,62 +49,66 @@ var navigatorFactory = function navigatorFactory(testContext, testMap) {
          * @param {String} direction - the move direction (next, previous or jump)
          * @param {String} scope - the move scope (item, section, testPart)
          * @param {Number} [position] - the position in case of jump
-         * @returns {Object|Boolean} - false if we can't navigate, otherwise the result of the nav
+         * @returns {Promise} - the result of the nav
          */
         navigate: function navigate(direction, scope, position) {
-            var methodName = direction.toLowerCase() + scope.substr(0, 1).toUpperCase() + scope.substr(1).toLowerCase();
+            return new Promise(function(resolve, reject) {
```
jsconan: No need to wrap the promise here.

```javascript
const methodName = direction.toLowerCase() + scope.substr(0, 1).toUpperCase() + scope.substr(1).toLowerCase();
if (_.isFunction(this[methodName])) {
    return this[methodName](position, attempt);
}
return Promise.reject(new Error(`Unknown navigation method ${methodName}`));
```

src/navigator/navigator.js
```diff
@@ -49,62 +49,66 @@ var navigatorFactory = function navigatorFactory(testContext, testMap) {
...
 
         /**
          * Navigate to the given position
          * @param {Number} position - the position
-         * @returns {Object} the new test context
+         * @returns {Promise} the new test context
          */
         jumpItem: function jumpItem(position) {
-            return testContextBuilder.buildTestContextFromPosition(testContext, testMap, position);
+            var item = mapHelper.getItemAt(updatedMap, position);
```
jsconan: `updatedMap` does not exist

src/navigator/navigator.js
```diff
@@ -49,62 +49,66 @@ var navigatorFactory = function navigatorFactory(testContext, testMap) {
...
+            if (item && itemStore.has(item.id)) {
```
jsconan: There is no `itemStore` in standard navigation.

src/navigator/navigator.js
```diff
@@ -49,62 +49,66 @@ var navigatorFactory = function navigatorFactory(testContext, testMap) {
...
+                return itemStore.get(item.id)
+                    .then(itemFromStore => {
+                        const newTestContext = testContextBuilder.buildTestContextFromPosition(testContext, testMap, position, itemFromStore.attempt)
+                        return itemStore.update(newTestContext.itemIdentifier, 'attempt', newTestContext.attempt)
```
jsconan: Why do you need the item store?
The attempt could also come from the existing testContext

src/navigator/navigator.js
```diff
@@ -49,62 +49,66 @@ var navigatorFactory = function navigatorFactory(testContext, testMap) {
...
+                            .then(() => newTestContext);
+                    })
+            }
+            return Promise.resolve(testContextBuilder.buildTestContextFromPosition(testContext, testMap, position));
```
jsconan: So the only reason why you need to return a promise is because of the itemStore?
In that case, you don't need it. There is no itemStore in the normal mode.

src/proxy/cache/proxy.js
```diff
@@ -134,56 +134,62 @@ export default _.defaults(
...
             this.offlineAction = function offlineAction(action, actionParams) {
-                var testNavigator;
-                var newTestContext;
-                var result = { success: true };
-
-                var blockingActions = ['exitTest', 'timeout'];
-
-                var testContext = this.getDataHolder().get('testContext');
-                var testMap = this.getDataHolder().get('testMap');
+                return new Promise(function(resolve, reject) {
```
jsconan: Is it required to return a promise?

src/helpers/testContextBuilder.js
```diff
@@ -87,14 +87,15 @@ function buildTestContextFromJump(testContext, testMap, jump) {
...
-function getTestContext(testContext, testMap, item, section, part, position) {
+function getTestContext(testContext, testMap, item, section, part, position, attempt) {
     var isLeavingSection = section.id !== testContext.sectionId,
         isLeavingPart = part.id !== testContext.testPartId,
         newTestContext = _.defaults(
             {
                 itemIdentifier: item.id,
                 itemPosition: position,
                 remainingAttempts: Math.max(-1, item.remainingAttempts - 1),
+                attempt: attempt ? attempt + 1 : 1,
```
jsconan: Maybe should you only set the attempt if provided. Otherwise, it is safer to not add it and keep the value from the test context

src/navigator/offlineNavigator.js
bziondik: fixed

src/helpers/testContextBuilder.js
bziondik: fixed

src/navigator/offlineNavigator.js
bziondik: fixed

PR #107, opened 2020-01-08
[TAO-9575] WTP-1413 - "time_remaining" in json is incorrect for exit_code 704

AntonTsymuk:
Related to : https://oat-sa.atlassian.net/browse/TAO-9575
 
Update item duration before sending it to server

#### How to test
 
- Prepare an instance of TAO with taoAct extension
- Prepare a delivery with multiple items and time limit
- Execute the delivery as a test taker
- Pass several question and wait until the end of time
- Make sure that in ODS Transmission Monitoring completedTestEvent has 0 time_remaining

### DIFF
--- a/src/plugins/controls/duration/duration.js
+++ b/src/plugins/controls/duration/duration.js
@@ -167,8 +167,6 @@ export default pluginFactory({
                             // the duration will be sent to the server with the next request,
                             // usually submitItem() or callItemAction()
                             testRunner.getProxy().addCallActionParams(params);
-
-                            self.disable();
                         });
                 })
 
--- a/src/plugins/controls/duration/duration.js
+++ b/src/plugins/controls/duration/duration.js
@@ -57,6 +57,26 @@ export default pluginFactory({
 
         //where the duration of attempts are stored
         return testRunner.getPluginStore(this.getName()).then(function(durationStore) {
+            let currentLock = Promise.resolve();
+
+            /**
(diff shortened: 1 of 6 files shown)


src/plugins/controls/duration/duration.js
```diff
@@ -70,66 +77,76 @@ export default pluginFactory({
...
-
                 .after('renderitem', function() {
                     self.enable();
                 })
                 .on('enableitem', function() {
                     self.enable();
                 })
-
+                .before('move skip exit timeout error disableitem', function() {
+                    updateDuration().then(() => self.disable());
```
krampstudio: could be good to catch it (if you don't want to report the error, at least log it)

src/plugins/controls/duration/duration.js
```diff
@@ -70,66 +77,76 @@ export default pluginFactory({
...
-                        };
-                        if (_.isNumber(duration) && duration > 0) {
-                            params.itemDuration = duration;
-                        }
-
-                        // the duration will be sent to the server with the next request,
-                        // usually submitItem() or callItemAction()
-                        testRunner.getProxy().addCallActionParams(params);
-                    });
+                    return currentUpdatePromise
```
krampstudio: In case of rejection the events chain will be interrupted. It could be better to catch errors and log them directly.
@jsconan what do you think ?

PR #149, opened 2020-03-17
Fix/actp 410/tools keyboard navigation

ampaveliev:
**Related to**
https://oat-sa.atlassian.net/browse/ACTP-410

**Description**
Sometimes it was't possible to navigate through tools menu via keyboard. This happened in case when first item was hidden.

**How to test**
1. Create a test with tools.
2. Start a test.
3. Try to navigate using up and down arrows.
4. Check different combination of tools.


src/ui/toolbox/menu.js
```diff
@@ -117,6 +117,30 @@ var menuComponentApi = {
...
     },
 
+    /**
+     * It needs to find nearest visible item.
+     *
+     * @param {-1|1} inc - incrementor. -1 - navigate to the top,  1 - to the bottom.
+     *
+     * @returns {JQuery} element.
+     */
+    _findClosestVisibleItem(inc = 1) {
```
jsconan: Private functions don't need to start by an underscore, see the [frontend coding guide](https://hub.taotesting.com/articles/frontend/architecture/tao-frontend-coding-guide#general-rules).

src/ui/toolbox/menu.js
```diff
@@ -117,6 +117,30 @@ var menuComponentApi = {
...
+        }
+        const last = inc === 1 ? this.menuItems.length : -1;
+        let elem;
+        do {
+            this.hoverIndex += inc;
+            if (this.hoverIndex === last) {
+                return;
+            }
+            elem = this.menuItems[this.hoverIndex].getElement();
+        } while (elem && elem.hasClass('hidden'));
```
jsconan: Why not rely on `lodash` for such an algorithm?

You migt have a clearer code using `_.find()` and `_.findLast()`.

Also you might simplify it by having 2 methods instead: `findFirstVisibleItem()` and `findLastVisibleItem()`. This will make the code more readable.

src/ui/toolbox/menu.js
```diff
@@ -148,16 +172,18 @@ var menuComponentApi = {
             // fromLast (default) navigation: focus on button and then using UP go to last item
             this.hoverIndex = this.menuItems.length - 1; // we start on the button, not at the max array index
             // which would be menuItems.length-1
-            if (this.$menuItems[this.hoverIndex]) {
-                this.$menuItems[this.hoverIndex].focus();
+            const elem = this._findClosestVisibleItem(-1);
+            if (elem) {
+                this.hoverItem(this.menuItems[this.hoverIndex].id);
```
jsconan: This code is not clear since `_findClosestVisibleItem()` is not a pure function. It is not obvious the function also modify the index.

A good implementation would be to have a function returning the index of the first visible elment instead, and then assign the value to the `hoverIndex` property.

src/ui/toolbox/menu.js
ampaveliev: Lodash methods are not fit, because we  rely on the hover index (it can be for instance 5). Also this function have to change hoverIndex, because menuItems can contain invisible items and we should skip them. Nonetheless I'll rename it to improve DX. 

Good point about 2 methods, I'll use currying

src/ui/toolbox/menu.js
ampaveliev: it just a helper to avoid code duplications. Probably we can return index and move updating hoverIndex to the hoverItem method , but this may affect existing code.

src/ui/toolbox/menu.js
ampaveliev: No problem, I'll remove it, but... Why? it's interesting to know the reason to mark a standard language notation as a bad practice.

jsconan: This is not a standard so far. The current state of the prefix notation for private members is still in stage 3, so not yet validated: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Class_fields#Private_fields

And it will not be the underscore.

You can also see some explanations here: https://stackoverflow.com/a/56205859

Moreover, we tend to not adopt too early new language features. For instance, we are still not allowing `async`/`await` yet. The reason why is we prefer to make sure all is properly supported and well mastered to avoid both too verbose transpilations and construction issues.

src/ui/toolbox/menu.js
jsconan: But this is exactly one of the flaws of the function: it has a side effect.
A better implementation would be to only compute the index, without modifying it.
Then the consumer that will call this function will have to update actually the index.

src/ui/toolbox/menu.js
jsconan: why do you want to update the `hoverItem` method? Can't you do something like this?:
```javascript
const index = this.findFirstVisibleItem();
if (index >= 0) {`
    this.hoverIndex = index;
    this.hoverItem(this.menuItems[this.hoverIndex].id);
}
```

src/ui/toolbox/menu.js
ampaveliev: I didn't mean language feature proposal, because to be truthy I hadn't heard about it. It's old javascript convention. An underscore before a name is used to specify that the name is to be treated as “private” by a programmer. 

From the second link:
> _ prefixed variable names are considered private by convention but are still public.

src/ui/toolbox/menu.js
ampaveliev: Most of the methods of this object are not pure, so I'm not sure that I can understand, whats wrong just with this one. Developer can easily understand by the name of the function that it applies some necessary side effects.

Anyway of course it's possible to split function into two: **hoverClosestVisibleItem** and **findClosestVisibleItem** the last one will be called by the first one. Is it ok?

src/ui/toolbox/menu.js
ampaveliev: Ok. if you insist, I'll do it as you suggest. But this code snippet will be duplicated in several places. This is one of the thing I tried to avoid, when moved the reusable code to the method.

src/ui/toolbox/menu.js
jsconan: Yes, unfortunately, the existing code might not follow the best practices. However, the goal is to avoid continuing the mess. So if the fix requires some refactoring, don't hesitate to do so.

PR #150, opened 2020-03-18
Feature/actp 297/refactor the keynavigation plugin

jsconan:
Related to: https://oat-sa.atlassian.net/browse/ACTP-297

Refactoring of the `keyNavigation` plugin. 
Unit tests have been added to secure the features before applying the changes. You can find a playground in the plugin's test. It exposes the 3 available modes.

image

The plugin has already been split down into the engine and the plugin itself (cf. https://github.com/oat-sa/tao-test-runner-qti-fe/pull/139). This PR goes a bit deeper by exploding the engine into separated `strategies`. Strategies are grouped by the config of a `mode`.

Strategies and modes are managed through a local registry, allowing to replace them at initialization time. For instance, a customer's plugin might register another mode with different strategies before the test starts.

Since the mode is declaring the strategies in an array, the order matters, and each mode can configure a particular screen navigation order, giving more flexibility for UX.

Note: some issues have been discovered with the native mode when the access to the page and to the rubrics are enabled. For this reason this access has not been added yet for the native mode, it will be the object of another PR that will fix the issue.

The plugin's behavior remains the same as before the refactoring, except for a navigation issue that has been fixed in the linear mode (the navigation was broken if the item contains more than 2 choices per lines). Some behavior has also been factorized to keep the feature as DRY as possible.

Added/Impacted unit tests:
- `test/plugins/content/keyNavigation/plugin/test.html`
- `test/plugins/content/keyNavigation/keyNavigation/test.html`
- `test/plugins/content/keyNavigation/helpers/test.html`
- `test/plugins/content/keyNavigation/modesManager/test.html`
- `test/plugins/content/keyNavigation/strategiesManager/test.html`

To run tests:
```
npm i
npm run test
```

You may scope to the `keyNavigation`: `npm run test keyNavigation`
Or run the test into the browser: `npm run test:keepAlive`


src/plugins/content/accessibility/keyNavigation/keyNavigation.js
```diff
@@ -0,0 +1,161 @@
...
+            return testRunner;
+        },
+
+        /**
+         * Switches the navigation mode
+         * @param {String} newMode
+         * @returns {testRunnerKeyNavigator}
+         */
+        setMode(newMode) {
+            contentNavigatorType = newMode;
```
bziondik: is it enough to switch mode? Or some part of function from `init` should be reinvoke

src/plugins/content/accessibility/keyNavigation/modes/nativeMode.js
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Builds the key navigation config for the "native" mode
+     * @param {keyNavigationStrategyConfig} config - additional config to set
+     * @returns {keyNavigationMode}
+     */
+    init(config = {}) {
+        return {
+            // todo: add access to the page and the rubric blocks
+            strategies: ['header', 'navigator', 'item', 'toolbar'],
```
bziondik: when I run `test:keepAlive` for `test/plugins/content/keyNavigation/plugin/test.html` in mode `Native`  I didn't reach left navigation panel by the keyboard. 
Or strategy `'navigator' is not left navigation panel?

jsconan: The first item of the mock test is in a linear section, so the navigator is not interactive. If you move to the next item, you should be able to reach it. However, the visual style may not be obvious (I didn't change this).

src/plugins/content/accessibility/keyNavigation/keyNavigation.js
jsconan: Yes, this is a `soft` change. The strategies will be applied on the next call to `init`.

src/plugins/content/accessibility/keyNavigation/helpers.js
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Checks whether element is navigable from
+ *
+ * @param {HTMLElement|keyNavigator} element
+ * @returns {boolean}
+ */
+export function allowedToNavigateFrom(element) {
+    if (element.getCursor) {
+        const {navigable} = element.getCursor();
+        element = navigable;
```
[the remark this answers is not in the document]
jsconan: You're right, this might introduce an issue in complex functions.
This parameter reassignment was introduced to fix an issue within the helper, and I chose to not use a local value to simplify the code. Anyway, I'll update it to prevent further issues if someone add more complexity in it later.

PR #158, opened 2020-03-27
Feature/actp 299/top action bar in page structure

AnastasiaArcadia:
**Related to:**
https://oat-sa.atlassian.net/browse/ACTP-299

**Description:**
Include the "Top Action Bar" into the page structure, so that it's content would be announced and will follow the logical structure of the page (after the "Exit" button)

**How to check:**
Login as test taker and start some test.
The top bar should be reachable so that the screen reader can announce its content.
If the review panel is enabled test time should be announced, otherwise not (see ticket comments for details)


src/plugins/controls/timer/plugin.js
```diff
@@ -177,6 +193,7 @@ export default pluginFactory({
                     .then(function(startZen) {
                         //set up the timerbox
                         self.timerbox = timerboxFactory({
+                            ariaHidden: isReviewPanelEnabled(),
```
jsconan: This does not work if the review panel is enabled later.
You should also do something  after `renderitem`: https://github.com/oat-sa/tao-test-runner-qti-fe/pull/158/files#diff-80a5ac2a02cffb57a0fcff319627f957R181

For instance, naive approach: `self.timerbox.getElement().attr('aria-hidden', isReviewPanelEnabled());`

test/plugins/content/keyNavigation/assets/tao-3.css
```diff
@@ -1 +1,3 @@
...ontainer-footer .square,.section-container .content-block .data-container-wrapper .data-container>footer .square,.section-container .content-block .data-container-wrapper .data-container .data-container-footer .square{width:28px}.section-container .content-block .data-container-wrapper>section>footer .square span,.section-container .content-block .data-container-wrapper>section .data-container-footer .square span,.section-container .content-block .data-container-wrapper .data-container>footer .square span,.section-container .content-block .data-container-wrapper .data-container .data-container-footer .square span{padding:0;left:0}.section-container .content-block .data-container-wrapper>section ol,.section-container .content-block .data-container-wrapper .data-container ol{margin:0 0 0 15px;padding:10px}.section-container .content-block #form-container.ui-widget-content{border:none !important}.section-container .content-block form:not(.list-container){border:1px #ddd solid;background:#f3f1ef;padding:30px;border:1px solid #ddd;border-radius:2px;-webkit-border-radius:2px}.section-container .content-block [class^="btn-"],.section-container .content-block [class*=" btn-"]{margin:0 2px}
```
ampaveliev, line 1: Don't see corresponding scss changes

ampaveliev:
The feature not working correctly for me. I can't see the focus when it is on the topbar. Also screenreaders just announce  "blank" when the focus is on the topbar. Have I missed something?

test/plugins/content/keyNavigation/assets/layout.tpl
```diff
@@ -1,40 +1,43 @@
...
-            <div class="settings-menu">
-                <ul class="clearfix plain">
-                    <li data-control="home">
-                        <a id="home" href="#">
-                            <span class="icon-home"></span>
-                            <span class="text hidden">Tests Home</span>
-                        </a>
-                    </li>
-                    <li class="infoControl sep-before">
+    <div role="banner" aria-label="Main Menu.">
```
ampaveliev, line 2: Should `Main menu` message be translatable?

src/plugins/controls/timer/component/tpl/timerbox.tpl
```diff
@@ -1,4 +1,4 @@
-<div class="timer-box">
+<div class="timer-box" aria-hidden="{{#if ariaHidden}}true{{else}}false{{/if}}">
```
ampaveliev, line 1: Can true/false be passed to the template instead?

ampaveliev:
The feature works for me, but I think that focused element on the topbar should be visually highlighted. Also should just hide / show timer be announced by screenreaders? I think it'd better to notify user , how much time is left

PR #168, opened 2020-04-13
Fix/tca 439/menu announce multiple times

AnastasiaArcadia:
**Related to**
https://oat-sa.atlassian.net/browse/TCA-439

**Description**
Selected option and menu title is announced multiple times for Tools and Contrast

**How to test**
1. As a test-taker start a test
2. Navigate to the bottom bar, set focus on "Contrast" button
3. Press Space / Enter to expand the list of options
4. JAWS should announce just "listbox - Black on white checked ..." without repetatives

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-test-runner-qti",
-    "version": "1.3.0",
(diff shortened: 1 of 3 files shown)


AnastasiaArcadia:
Rerun npm install, is it better now?

PR #181, opened 2020-04-26
Fix/tca 483/review tabs incorrectly announced

AnastasiaArcadia:
**Related to**
https://oat-sa.atlassian.net/browse/TCA-483

**Description**
KB actions for All / Unanswered / Flagged tabs are incorrectly announced

**How to test**
1. as a user start a test
2. enable JAWS
3. navigate to All / Unanswered / Flagged tabs
4. set focus to each tab, listen to SR's announcements

It should be no announcement of navigation via Arrow keys

### DIFF
--- a/src/plugins/content/accessibility/keyNavigation/strategies/navigatorNavigation.js
+++ b/src/plugins/content/accessibility/keyNavigation/strategies/navigatorNavigation.js
@@ -63,7 +63,13 @@ export default {
                     propagateTab: false,
                     elements: navigableFilters,
                     group: $navigator.find('.qti-navigator-filters')
-                });
+                })
+                    .on('focus', cursor => {
+                        cursor.navigable.getElement().attr('aria-selected', true);
+                    })
+                    .on('blur', cursor => {
+                        cursor.navigable.getElement().attr('aria-selected', false);
+                    });
 
                 setupItemsNavigator(filtersNavigator, {
                     keyNextItem: config.keyNextTab || config.keyNextItem,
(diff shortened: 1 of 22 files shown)


src/plugins/content/accessibility/keyNavigation/strategies/navigatorNavigation.js
```diff
@@ -63,7 +63,13 @@ export default {
                     propagateTab: false,
                     elements: navigableFilters,
                     group: $navigator.find('.qti-navigator-filters')
-                });
+                })
+                    .on('focus', cursor => {
```
ampaveliev, line 67: I guess the key navigator should not care about handling events related to target elements. It should be their own responsibility.

AnastasiaArcadia: But we already have this approach here
https://github.com/oat-sa/tao-test-runner-qti-fe/blob/f28f17b3158531668365238c3f2e05f27a99a6f3/src/plugins/content/accessibility/keyNavigation/strategies/navigatorNavigation.js#L75
https://github.com/oat-sa/tao-test-runner-qti-fe/blob/f28f17b3158531668365238c3f2e05f27a99a6f3/src/plugins/content/accessibility/keyNavigation/strategies/navigatorNavigation.js#L139

ampaveliev: The handlers in your examples are related to the key navigator behavior itself. Anyway you should keep target elements logic in the  appropriate modules, if it possible

PR #185, opened 2020-04-30
tts plugin keep across items TCA-381

lecosson:
**Related to**
https://oat-sa.atlassian.net/browse/TCA-381

**Description**
test taker wants to have the TTS tool defaulted to on when I begin the test and stay on (unless turn it off) across all items of the test.

**How to test**
1. prepare ACT instance
2. link or replace package with this branch
3. navigate to test with some APIP items
4. check APIP plugin state and position across items

**Related PR's**
https://github.com/oat-sa/extension-tao-testqti/pull/1767

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-test-runner-qti",
-    "version": "2.1.2",
+    "version": "2.1.3",
     "description": "TAO Test Runner QTI implementation",
     "files": [
         "dist",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-test-runner-qti",
-    "version": "2.1.1",
+    "version": "2.1.2",
     "description": "TAO Test Runner QTI implementation",
     "files": [
         "dist",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-test-runner-qti",
-    "version": "2.1.3",
+    "version": "2.1.4",
     "description": "TAO Test Runner QTI implementation",
     "files": [
         "dist",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-test-runner-qti",
-    "version": "2.1.0",
+    "version": "2.1.1",
     "description": "TAO Test Runner QTI implementation",
     "files": [
         "dist",
--- a/src/plugins/tools/apipTextToSpeech/plugin.js
+++ b/src/plugins/tools/apipTextToSpeech/plugin.js
@@ -204,6 +204,7 @@ export default pluginFactory({
             if (this.getState('enabled')) {
                 if (this.getState('active')) {
                     disablePlugin();
+                    this.setState('sleep', true);
                 } else {
                     enablePlugin();
                 }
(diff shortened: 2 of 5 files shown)


src/plugins/tools/apipTextToSpeech/plugin.js
```diff
@@ -316,6 +318,10 @@ export default pluginFactory({
 
                 getTTSComponent().setMediaContentData(ttsApipData);
                 this.show();
+                if (!this.getState('sleep')) {
+                    this.setState('enabled', true);
```
jsconan, line 322: Why not use `this.enable()` instead? The state `enabled` is managed internally by the methods `enable()` and `disable()` (cf. [the plugin abstraction](https://github.com/oat-sa/tao-core-sdk-fe/blob/8cf13ba0c121d845d44c6e675084920cabacf6ef/src/core/plugin.js#L325:L344)).

ampaveliev:
Looks ok, but I can't test it locally. Because I don't understand how should I prepare my local instance and what kind of items are called APIP:

> 1. prepare ACT instance
> 2. ...
>  3. navigate to test with some APIP items

src/plugins/tools/apipTextToSpeech/plugin.js
```diff
@@ -224,6 +225,7 @@ export default pluginFactory({
         // Handle plugin button click
         this.button.on('click', (e) => {
             e.preventDefault();
+            this.setState('sleep', false);
```
jsconan, line 228: Only when clicking on the button? Maybe could it be better to have this when doing the action instead or am I mistaken? (when catching the event)

lecosson:
> Looks ok, but I can't test it locally. Because I don't understand how should I prepare my local instance and what kind of items are called APIP:
> 
> > 1. prepare ACT instance
> > 2. ...
> > 3. navigate to test with some APIP items

text example added to jira

src/plugins/tools/apipTextToSpeech/plugin.js
lecosson: there is `enable` method relates to button, not to plugin

src/plugins/tools/apipTextToSpeech/plugin.js
lecosson: yes, has a reason, fixed

lecosson:
> Code review only.
> Could you add the package-lock file ?
> 
> Unfortunately the unit tests are failing (timeout on CLI, and error message in browser)
> image

fixed

src/plugins/tools/apipTextToSpeech/plugin.js
```diff
@@ -316,6 +318,10 @@ export default pluginFactory({
 
                 getTTSComponent().setMediaContentData(ttsApipData);
                 this.show();
+                if (!this.getState('sleep')) {
+                    this.setState('enabled', true);
+                    toggleTool();
```
jsconan, line 323: This function is calling `enablePlugin()`, which is recreating the keyNavigation. This leads to inconsistency and make the unit tests failing.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-test-runner-qti",
-    "version": "2.1.1",
+    "version": "2.1.3",
```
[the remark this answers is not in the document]
lecosson: 2.1.2 already released

src/plugins/tools/apipTextToSpeech/plugin.js
lecosson: line added to test, test is passed

PR #194, opened 2020-05-10
Feature/tca 307/review panel expose status

AnastasiaArcadia:
Related to : https://oat-sa.atlassian.net/browse/TCA-307
 
TestRunner - ReviewPanel - Expose the status of parts, sections and items.
  
#### How to test

- Create a test with the review panel enabled
- Start a test as a test taker
- Navigate to the review panel via Tab
- Check that screen reader announces:
  - the label of the test part (if present)
  - the label of the test section (if present)
  - the position of the focused item
  - the state of the focused item (answered, viewed, etc.)
  - that it is a link
  - that it is current
  - the label of the item

(Keyboard navigation issues are not taken into account here)

### DIFF
--- a/src/plugins/navigation/review/navigatorTree.tpl
+++ b/src/plugins/navigation/review/navigatorTree.tpl
@@ -33,7 +33,7 @@
                         <span class="qti-navigator-label truncate" title="{{label}}"
                               role="link" aria-disabled="{{#if viewed}}false{{else}}true{{/if}}"
                               {{#if active}}aria-current="page"{{/if}}
-                              aria-label="{{index}} of {{../stats.questions}} {{icon}}">
+                              aria-label="{{index}} of {{../stats.total}} {{icon}}">
                             <span class="qti-navigator-icon icon-{{icon}}"></span>
                             <span class="qti-navigator-number">{{index}}</span>
                             {{label}}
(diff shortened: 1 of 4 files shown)


src/plugins/navigation/review/navigatorTree.tpl
```diff
@@ -20,17 +20,20 @@
...
                 </span>
-                <ul class="qti-navigator-items collapsible-panel plain">
+                <ul aria-label="{{label}}" class="qti-navigator-items collapsible-panel plain">
                     {{#each items}}
                     <li class="qti-navigator-item {{cls}}" data-id="{{id}}" data-position="{{position}}">
-                        <span class="qti-navigator-label truncate" title="{{label}}">
+                        <span class="qti-navigator-label truncate" title="{{label}}"
+                              role="link" aria-disabled="{{#if viewed}}false{{else}}true{{/if}}"
+                              {{#if active}}aria-current="page"{{/if}}
+                              aria-label="{{index}} of {{../stats.questions}} {{icon}}">
```
jsconan, line 36: This will only work for actual questions. Informational items won't be taken into account.
You should rather use `stats.total` instead of `stats.questions`. Otherwise, the number of presented items won't be reflected by the screen reader announcement.

PR #199, opened 2020-05-13
[TCA-501] TestRunner - Passages - Ability to scroll item content using KB

atsymuk:
Related to : https://oat-sa.atlassian.net/browse/TCA-501
 
Add stimulus navigation strategy
  
#### How to test
 
- prepare an item with shared stimulus(item can be found in the comment to the ticket)
- prepare a delivery with the item
- Execute a delivery as a test taker
- Try to navigate to the shared stimulus using keyboard navigation and scroll it's content
  
#### Dependencies
   
Companion PR :
 - [ ] https://github.com/oat-sa/extension-tao-testqti/pull/1780

### DIFF
--- a/src/plugins/content/accessibility/keyNavigation/strategies/stimulusNavigation.js
+++ b/src/plugins/content/accessibility/keyNavigation/strategies/stimulusNavigation.js
@@ -1,0 +1,91 @@
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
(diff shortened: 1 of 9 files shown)


src/plugins/content/accessibility/keyNavigation/strategies/stimulusNavigation.js
```diff
@@ -0,0 +1,91 @@
...
+    init() {
+        const config = this.getConfig();
+        this.keyNavigators = [];
+
+        const $content = this.getTestRunner().getAreaBroker().getContentArea();
+        $content
+            .find('.stimulus-container')
+            // filter out nodes without scrollbar
+            .filter(function () {
+                return this.scrollHeight > this.clientHeight;
```
jsconan, line 54: Well, this might work if the size is set before the keyNavigator is created. Unfortunately this is rarely the case. 
A better approach would be to target all stimulus, then dynamically compute if the element can be focused.

A possibility would be to overload the `isEnabled()` method in the `navigableDomElement` instance when creating the keyNavigator:
```javascript
                const elements = navigableDomElement.createFromDoms($element).map(el => {
                    const oldIsEnabled = el.isEnabled;
                    el.isEnabled = function isEnabled() {
                        if (oldIsEnabled.call(this)) {
                            const node = this.getElement().get(0);
                            return node.scrollHeight > node.clientHeight;
                        }
                        return false;
                    };
                    return el;
                });
                const navigator = keyNavigator({
                    id: `${groupId}-${i}`,
                    elements,
                    group: $element,
                    propagateTab: false
                });
```

lecosson:
> Companion PR

missed links

PR #201, opened 2020-05-13
Fix/tca 552/review panel jumplinks key navigator

ampaveliev:
_CASE 1_
**Related to:**
https://oat-sa.atlassian.net/browse/TCA-552

**Description**
Test status and structure" jump link sets focus on "Test status" button, but next TAB stop is on Exit button

**How to test**
_Preconditions:_
create a test with Review panel (Enable review scree, Enable mark for review test navigation settings are on)

_Steps:_

1. start the test from preconditions
2. hit TAB button to navigate to jump links
3. select "Test status and structure" jump link, hit Enter
4. navigate inside Review panel using KB

_CASE 2_
**Related to**
https://oat-sa.atlassian.net/browse/TCA-456#icft=TCA-456

**Description**
SR user can reach "Test status menu" only when scanning the page, although it's an interactive element and should have a TAB stop

**How to test**
_Preconditions:_
create a Delivery with "Review panel" enabled

_Steps:_

1. as a user start a test
2. using TAB navigate through the interactive elements on the page
3. change key navigator plugin strategy and test again

### DIFF
--- a/src/plugins/content/accessibility/keyNavigation/strategies/navigatorNavigation.js
+++ b/src/plugins/content/accessibility/keyNavigation/strategies/navigatorNavigation.js
@@ -53,7 +53,47 @@ export default {
         this.managedNavigators = [];
         this.keyNavigators = [];
 
+        let testStatusNavigation;
         if ($navigator.length && !$navigator.hasClass('disabled')) {
+            const $testStatusHeader = $navigator.find('.qti-navigator-info.collapsible .qti-navigator-header');
+            const navigableTestStatus = navigableDomElement.createFromDoms($testStatusHeader);
+
+            if (navigableTestStatus.length) {
+                testStatusNavigation = keyNavigator({
+                    keepState: config.keepState,
+                    id: 'navigator-test-status',
+                    propagateTab: false,
+                    elements: navigableTestStatus,
+                    group: $testStatusHeader,
+                });
+
+                setupItemsNavigator(testStatusNavigation, {
+                    keyNextItem: config.keyNextTab || config.keyNextItem,
+                    keyPrevItem: config.keyPrevTab || config.keyPrevItem
+                });
+                setupClickableNavigator(testStatusNavigation);
+
+                if (config.keyNextContent) {
+                    testStatusNavigation.on(config.keyNextContent, elem => {
+                        if (allowedToNavigateFrom(elem) && filtersNavigator) {
+                            filtersNavigator.focus();
+                        }
+                    });
+                }
+                if (config.keyPrevContent) {
+                    testStatusNavigation.on(config.keyPrevContent, elem => {
+                        if (allowedToNavigateFrom(elem) && itemsNavigator) {
+                            _.defer(() => {
+                                itemsNavigator.last();
+                            });
+                        }
+                    });
+                }
+
+                this.keyNavigators.push(testStatusNavigation);
+                this.managedNavigators.push(testStatusNavigation);
+            }
+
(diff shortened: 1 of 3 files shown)


src/plugins/content/accessibility/keyNavigation/strategies/navigatorNavigation.js
```diff
@@ -53,7 +53,47 @@ export default {
...
+                    propagateTab: false,
+                    elements: navigableTestStatus,
+                    group: $testStatusHeader,
+                });
+
+                setupItemsNavigator(testStatusNavigation, {
+                    keyNextItem: config.keyNextTab || config.keyNextItem,
+                    keyPrevItem: config.keyPrevTab || config.keyPrevItem
+                });
+                setupClickableNavigator(testStatusNavigation);
```
jsconan, line 74: No need to capture the activation of the element. It won't prevent the panel to be collapsed if you remove this line, nor it will add something if you keep it.

src/plugins/content/accessibility/keyNavigation/strategies/navigatorNavigation.js
```diff
@@ -53,7 +53,47 @@ export default {
...
+
+                if (config.keyNextContent) {
```
jsconan, line 76: Is it required to have the status accessible using arrow keys?
I would rather rely on the tab key only. Moreover, this is introducing an issue in 'default' mode when the focus is on the list of items: the 'right' key gives back the focus to the status instead of moving to the next tab.
I'd suggest to remove these lines.

src/plugins/content/accessibility/keyNavigation/strategies/navigatorNavigation.js
```diff
@@ -102,15 +142,19 @@ export default {
...
                             _.defer(() => {
                                 itemsNavigator.last();
-                            });
+                            })
                         }
                     });
                 }
 
-                this.keyNavigators.push(filtersNavigator);
+                if (!testStatusNavigation || !config.keyNextTab || !config.keyPrevTab) {
```
jsconan, line 155: This is preventing the filters to be accessed using the tab key in linear mode.

ampaveliev: It's ok. In linear mode the user uses tab to travel between strategies (header, topbar, review panel etc) and he should use arrows to switch between elements inside the group. So in this case the filters is accessible by tab if `keyNextTab `and `keyPrevTab` are defined and no `testStatusNavigation `element inside the review panel.

src/plugins/content/accessibility/keyNavigation/strategies/navigatorNavigation.js
```diff
@@ -53,7 +53,47 @@ export default {
...
+                    id: 'navigator-test-status',
+                    propagateTab: false,
+                    elements: navigableTestStatus,
+                    group: $testStatusHeader,
+                });
+
+                setupItemsNavigator(testStatusNavigation, {
+                    keyNextItem: config.keyNextTab || config.keyNextItem,
+                    keyPrevItem: config.keyPrevTab || config.keyPrevItem
+                });
```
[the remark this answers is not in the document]
ampaveliev: Test status becomes a trap in native mode without it.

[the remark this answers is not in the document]
jsconan: ok, got it. However, it should work once TCA-557 has been merged. Anyway, this is not a big deal to keep this.

PR #202, opened 2020-05-14
[TCA-521] Timer Correction - Introduce new central timer

atsymuk:
Related to : https://oat-sa.atlassian.net/browse/TCA-521
 
Introduce ф central timer inside test runner provider. 
The timer is client based and its responsibility is to counts out the time spent on the test.
The timer continues counting the time only when the test is interactive otherwise it is paused.
The central timer can be used by subscription to `tick` test runner event.
  
#### How to test
 
**Note:** Currently the timer is not used anywhere, but you still can make some change to subscribe to the test runner `tick` event to check if it is working

`testRunner.on('tick', function(...args) { console.log(args); })`

- prepare a delivery
- make some change to subscribe to `tick` test runner event
- execute the delivery as a test taker and check that event is triggered

Requires :
 - [ ] https://github.com/oat-sa/extension-tao-testqti/pull/1783

### DIFF
--- a/src/provider/stopwatch.js
+++ b/src/provider/stopwatch.js
@@ -34,54 +34,69 @@ const defaultOptions = {
  * @param {Number} [options.interval]
  * @returns {Object} stopwatch instance
  */
-export default function stopwatchFactory(options) {
+export default function stopwatchFactory(options = {}) {
     const config = Object.assign({}, defaultOptions, options);
+    let initialized = false;
+    let polling;
+    let stopwatch;
 
     return eventifier({
+        /**
+         * Is this instance initialized
+         * @returns {Boolean}
+         */
+        isInitialized() {
+            return initialized;
+        },
         /**
          * Initialize stopwatch
          */
         init() {
-            this.stopwatch = timerFactory({
+            stopwatch = timerFactory({
                 autoStart: false,
             });
 
             /**
              * @fires tick - every time when interval is elapsed
              */
-            this.polling = pollingFactory({
-                action: () => this.trigger('tick', this.stopwatch.tick()),
+            polling = pollingFactory({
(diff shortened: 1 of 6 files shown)


src/provider/stopwatch.js
```diff
@@ -0,0 +1,88 @@
...
+            if (this.initialized) {
+                this.stopwatch.pause();
+                this.polling.stop();
+            }
+        },
+        /**
+         * Destory stopwatch by stoping the timer
+         */
+        destroy() {
+            if (this.initialized) {
```
jsconan, line 82: Since this is a destroy step, you should also free the resources (destroy and null `stopwatch` and `polling`), then reset the state (`initialized` to `false`). Otherwise, nothing prevents to continue using the instance.

src/provider/stopwatch.js
```diff
@@ -0,0 +1,88 @@
...
+            /**
+             * @fires tick - every time when interval is elapsed
+             */
+            this.polling = pollingFactory({
+                action: () => this.trigger('tick', this.stopwatch.tick()),
+                interval: config.interval,
+                autoStart: false,
+            });
+
+            this.initialized = true;
```
jsconan, line 58: It would be better to use local variables to store the state (`stopwatch`, `polling`, `initialized`). The API should be the only one mean to manipulate the instance state. Otherwise, it won't prevent the outside world from interfering with it.

test/provider/stopwatch/test.html
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="utf-8" />
+        <title>Test Runner - QTI Provider</title>
```
jsconan, line 5: Wrong title

PR #209, opened 2020-05-19
Feature/tca 557/improve key navigation for radio buttons and review panel

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TCA-557

Requires: 
 - [ ] https://github.com/oat-sa/tao-core-ui-fe/pull/125

Refine the key navigation:
- in native key navigation mode, use arrow keys for radio choices and list of items, use tabs for other navigation
- auto-focus the selected radio button when reaching an interaction using the key navigation
- select the focused radio button when using the key navigation in a choice interaction
- allows navigating to unseen items using the keyboard in the navigation panel when the skip ahead option is set

With respect to the ticket, the following acceptance criteria are not met, and should be addressed later:
- Hitting TAB outside of the radio button group sets focus on:
    - the first radio button, if the user is using TAB and no radio button is selected
    - the last radio button, if the user is using shift+TAB and no radio button is selected
    - the radio button must be focused but no selected if no button was selected before moving to the group (the next move inside the group will change the focus **and** select the button)

To test the feature you need to:
- create some items with choice interactions both with checkboxes and radio buttons
- create test with these items
- enable the review panel as well the skip ahead option
- publish the delivery
- enable the native mode
- pass the test:
    - the choices with checkboxes must still be navigable using the TAB key
    - the choices with radio buttons must now be navigable using the ARROW keys
    - the navigation panel must be navigable using TAB key to move from the filters to the list of items, inside filters or items the navigation must be made with ARROW keys

**Warning**: the companion PR must be merged before, then don't forget to update the `package.json` file with the proper version


src/plugins/content/accessibility/keyNavigation/strategies/itemNavigation.js
```diff
@@ -43,75 +43,126 @@ export default {
...
-                //filter out interaction as it will be managed separately
-                return !$(this).parents('.qti-interaction').length;
+            .filter((i, node) => {
+                const $node = $(node);
+                // reduce the list to full interactions, each one will get its own keyNavigator
+                if (!$node.parents('.qti-interaction').length) {
+                    const $choiceInput = $node.find('.qti-choice input');
+                    if ($choiceInput.length) {
+                        // expose the type of choices: checkbox or radio
+                        $node.attr('data-choice-type', $choiceInput.attr('type'));
```
ampaveliev, line 109: Filter method should accept pure function as a callback. I suggest to use `each` for applying side-effects

ampaveliev:
I feel discomfort because I see particular logic related to the specific areas in the global plugin. In my opinion, such plugins should be pretty simple and straightforward. Can this logic (like exposing choice interaction type and handling concrete keyboard keys) be moved to the particular places? Or can such logic be managed by the concrete plugins itself?

jsconan:
Well, I see your point. Eventually, the input type detection might be moved into choice interaction. However, this is not needed anymore, the type is not written anymore in the dom, the detection is now fully internal.
Regarding the key handling, it is already placed at a particular place, as this is the responsibility of the itemNavigation strategy to manage that. A better approach would be to get rid of the keyNavigation plugin, but this is another debate, not applicable for the current version of TAO.

PR #216, opened 2020-05-21
[TCA-603] Timer information (static, non-changing) is announced in the top bar

atsymuk:
Related to : https://oat-sa.atlassian.net/browse/TCA-603
 
Clean up screenreader timer notification after timeout or navigation between items
  
#### How to test
 First scenario: 
- prepare a delivery with a section timer set to 3 minutes
- execute a delivery as a test taker
- wait for screen reader notification about remaining time which will be triggered when timer reaches 2 minutes
- make sure that notification cleaned up after 20 seconds

Second scenario:
- prepare a delivery with a section timer set to 3 minutes
- execute a delivery as a test taker
- wait for screen reader notification about remaining time which will be triggered when timer reaches 2 minutes
- navigate to the next item 
- make sure that notification cleaned up
  
#### Dependencies
   
Companion PR :
 - [ ] https://github.com/oat-sa/<EXT>/pull/<ID>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-test-runner-qti",
-    "version": "2.5.0",
+    "version": "2.5.1",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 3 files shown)


ampaveliev:
In my case notification was cleaned after 5 seconds. Is it ok?

PR #231, opened 2020-06-01
[TCA-590] Adjust the Keyboard Navigation overlay with correct Information

atsymuk:
Related to : https://oat-sa.atlassian.net/browse/TCA-590
 
Refactor jumplinks shortcuts to make it accessibility friendly
  
#### How to test
 
- prepare a delivery with enabled accessibility feature
- execute delivery as a test taker
- activate shortcuts jump link
- check AC from related ticket
  
#### Dependencies

Companion PR :
 - [ ] https://github.com/oat-sa/extension-tao-act/pull/1386
 - [ ] https://github.com/oat-sa/extension-tao-testqti/pull/1809

### DIFF
--- a/src/plugins/content/accessibility/jumplinks/shortcuts.js
+++ b/src/plugins/content/accessibility/jumplinks/shortcuts.js
@@ -21,14 +21,46 @@
  *
  * @author aliaksandr paliakou <lecosson@gmail.com>
  */
+import __ from 'i18n';
 import _ from 'lodash';
 import component from 'ui/component';
+import keyNavigator from 'ui/keyNavigation/navigator';
+import navigableDomElement from 'ui/keyNavigation/navigableDomElement';
 import shortcutsTpl from 'taoQtiTest/runner/plugins/content/accessibility/jumplinks/shortcuts.tpl';
 
 /**
  * Default config values, see below.
  */
-const defaults = {};
+const defaults = {
+    shortcutsGroups: [
+        {
+            id: 'navigation-shortcuts',
+            label: __('Navigation shortcuts'),
+            shortcuts: [
+                {
+                    id: 'next',
+                    shortcut: 'ALT + Shift + N',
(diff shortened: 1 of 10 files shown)


src/plugins/content/accessibility/jumplinks/shortcuts.js
```diff
@@ -39,14 +71,55 @@ const defaults = {};
  * @returns {shortcutsBox} the component, initialized and rendered
  */
 export default function shortcutsBoxFactory(config) {
+    const shortcutsBox = component({}, defaults)
+        .on('render', function () {
```
ampaveliev, line 75: Why didn't you use the dialog component? Seems it is doing most of the work.

atsymuk: I tried to use dialog component, but there was problems with screenreader announcement of the shortcuts, so we decided to stay with it but make it looks like dialog

atsymuk:
@art

> Works fine, but not the all AC are met
> 
> > The shortcuts should be presented using table markup (<table>, <tbody>, <tr>, <td>, …).
> 
> and the zooming section is missed. However I'm not sure is there any cross browser compatible API to apply zooming shortcuts, I think the appropriate comment should be added
> 
> * [x]   Tests are running successfully (old and new ones) on my local machine (if applicable)
> * [x]   New code is respecting code style rules
> * [x]   New code is respecting best practices
> * [x]   Feature is working correctly on my local machine (if applicable)
> * [ ]   Acceptance criteria are respected
> * [x]   Pull request title and description are meaningful

Since zooming behaviour is different across browsers, we decided to go without zooming shortcuts, but add some information about zooming to modal description

Also about the markup, we tried to use different variants:
- table
- definition list
- unordered list
- unordered list with ARIA attributes

All variants was reviewed by Luc and the unordered list was chosen as the best experience variant

I will add a comment about it to the ticket

PR #234, opened 2020-06-03
Feature/tca 596 timer synchronization

AnastasiaArcadia:
**Related**
https://oat-sa.atlassian.net/browse/TCA-596

**Description**
Timer synchronization between H1 and ReviewPanel Timer

**How to test**
1. Create test with timers
2. Start test as a test taker
3. Time in the review panel, top bar heading and timers should be consistent,
time should be announced as `%d hours, %d minutes, %d seconds to answer remaining %d questions (the current question)`

### DIFF
--- a/src/helpers/getTimerMessage.js
+++ b/src/helpers/getTimerMessage.js
@@ -1,0 +1,38 @@
+import __ from 'i18n';
+
+/**
+ * Returns timer information string
+ * @returns {Boolean}
+ */
+function getTimerMessage (hours, minutes, seconds, unansweredQuestions) {
+    let timerMessage;
+
+    const timeArr = [hours, minutes, seconds];
+    const timeArgArr = [];
+    [__('hours'), __('minutes'), __('seconds')].forEach((unit, idx) => {
+        if (timeArr[idx] > 0) {
+            timeArgArr.push(`${timeArr[idx]} ${unit}`);
+        }
+    });
+
+    let answeredMessage;
+    if (typeof unansweredQuestions !== 'number') {
+        answeredMessage = __('the current question');
+    } else {
+        let questionsMessage = __('questions');
+        if (unansweredQuestions === 1) {
+            questionsMessage = __('question');
+        }
+        answeredMessage = __('remaining %s %s', unansweredQuestions, questionsMessage);
+    }
+
+    if (timeArgArr.length === 0) {
+        timerMessage = __('%s to answer %s', 'no time left', answeredMessage);
+    } else {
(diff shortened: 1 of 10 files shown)


src/plugins/controls/title/title.js
```diff
@@ -144,26 +103,32 @@ export default pluginFactory({
...
                     || (unansweredQuestions && (unansweredQuestions !== currentUnansweredQuestions))
                 ) {
                     // update current timer state
-                    this.titles[scope].hours = hours;
-                    this.titles[scope].minutes = minutes;
-                    this.titles[scope].unansweredQuestions = unansweredQuestions;
-
-                    $timer.text(getTimerMessage(hours, minutes, unansweredQuestions));
+                    this.titles[scope] = {
+                        ...this.titles[scope],
```
atsymuk, line 125: `...` is not allowed by eslint configuration . please use `Object.assign` instead

src/plugins/controls/title/title.js
```diff
@@ -144,26 +103,32 @@ export default pluginFactory({
...
                     unansweredQuestions: currentUnansweredQuestions,
                 } = this.titles[scope];
                 const time = moment.duration(remainingTime / precision, 'seconds');
                 const hours = time.get('hours');
                 const minutes = time.get('minutes');
+                const seconds = time.get('seconds');
                 const unansweredQuestions = stats && (stats.questions - stats.answered);
 
                 // check if notification should be updated
                 if (
```
atsymuk, line 117: please remove this `if`, it is not relevant anymore since the `timertick` event triggered every second

src/helpers/getTimerMessage.js
```diff
@@ -0,0 +1,38 @@
+import __ from 'i18n';
+
+/**
+ * Returns timer information string
+ * @returns {Boolean}
+ */
+function getTimerMessage (hours, minutes, seconds, unansweredQuestions) {
```
atsymuk, line 7: please use arrow function

src/plugins/controls/timer/plugin.js
```diff
@@ -175,11 +187,14 @@ export default pluginFactory({
                             const timers = self.timerbox.getTimers();
 
                             const updatedTimers = Object.keys(timers).reduce((acc, timerName) => {
+                                const stats = statsHelper.getInstantStats(timers[timerName].scope, testRunner);
```
atsymuk, line 190: if scope is not supported by `getInstantStats`, it will return `test` scope. `item` scope is not supported, so, if you will try to get `stats` for `item` scope, it will return `test` scope stats. please consider it and generate proper message for `item` scope timer

PR #246, opened 2020-06-11
Fix/tca 647/remaining time mismatch

augustas:
#### Related
https://oat-sa.atlassian.net/browse/TCA-647

#### Description
`itemDuration` parameter was missing in `pause` request.

#### How to test

- Set up a customer environment with proctoring enabled and test with a time limit
- Pause the test session from proctor interface
- During next heartbeat the pause will be detected and additional pause request will be sent to BE. Verify that the `pause` request has `itemDuration` parameter.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "@oat-sa/tao-test-runner-qti",
-    "version": "2.10.1",
+    "version": "2.10.2",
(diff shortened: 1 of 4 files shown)


augustas:
> please remove `stopwatch.stop();` from exit callback as well

Thank you for spotting it, PR updated

PR #263, opened 2020-06-18
Fix NEX 1184 - Clear selected contrast on destroy

habemuscode:
**Related to**

https://oat-sa.atlassian.net/browse/NEX-1184

**Description**

When the user selects any contrast color in Preview and quits Preview mode, this selected color remains when the next Preview mode is opened, however when devices selector is change from default, every new Preview it is back to default.

**Steps to reproduce**

- Open an item (or shared stimulus) in Preview mode
- Change contrast color to 'White on Black'
- Quit Preview mode
- Open another item (or shared stimulus) in Preview mode

**Actual result:** Next Preview mode opened has the contrast color as was set previously.

**Expected result:** Every new opening of Preview mode has contrast color set to default 'Black on white'.

### DIFF
--- a/src/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
+++ b/src/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
@@ -260,7 +260,7 @@ export default pluginFactory({
      * Called during the runner's destroy phase
      */
     destroy: function destroy() {
-        this.init();
+        self.storage.setItem('itemThemeId', '');
         shortcut.remove(`.${this.getName()}`);
     },
 
--- a/src/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
+++ b/src/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
@@ -260,7 +260,7 @@ export default pluginFactory({
      * Called during the runner's destroy phase
      */
     destroy: function destroy() {
-        self.storage.setItem('itemThemeId', '');
+        this.storage.setItem('itemThemeId', '');
         shortcut.remove(`.${this.getName()}`);
     },
 
--- a/src/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
+++ b/src/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
@@ -259,6 +259,7 @@ export default pluginFactory({
      * Called during the runner's destroy phase
      */
     destroy: function destroy() {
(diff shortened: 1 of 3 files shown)


src/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
```diff
@@ -259,6 +260,7 @@ export default pluginFactory({
      * Called during the runner's destroy phase
      */
     destroy: function destroy() {
+        this.init();
```
wazelin, line 263: Could you please explain why it's needed? Can there be a simpler solution to resetting the values?

habemuscode: Well, in init function you have the reset, and for avoid refactor a lot of stuff, this is the simple way to go. You cannot access to any variables because are inside of init function instead at plugin level

habemuscode: And refactor more things, apart of go outside of scope of task, can add some bugs in future when regression tests that we dont found now

wazelin: I see your point.

If you think some refactoring is needed, could you please open a technical task to cover it?

habemuscode: https://oat-sa.atlassian.net/browse/TBD-36

habemuscode: But every file in currentTao needs refactor, I don't see the benefit for just refactor 1 file. Not benefit for us and not for customer

wazelin: It should be continuous work. That's why I initially suggested to apply a refactoring right away, boy scout rule.
Otherwise it has low chances of being prioritized.

src/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
```diff
@@ -259,6 +260,7 @@ export default pluginFactory({
      * Called during the runner's destroy phase
      */
     destroy: function destroy() {
+        self.storage.setItem('itemThemeId', '');
```
bziondik, line 263:
```suggestion
        this.storage.setItem('itemThemeId', '');
```

PR #274, opened 2020-06-30
Fix NEX 1184 - Clear selected contrast on destroy

habemuscode:
**PR related**

https://github.com/oat-sa/tao-test-runner-qti-fe/pull/263

**Issue**

https://oat-sa.atlassian.net/browse/NEX-1184

**Description**

When the user selects any contrast color in Preview and quits Preview mode, this selected color remains when the next Preview mode is opened, however when devices selector is change from default, every new Preview it is back to default.

**Steps to reproduce**

- Open an item (or shared stimulus) in Preview mode
- Change contrast color to 'White on Black'
- Quit Preview mode
- Open another item (or shared stimulus) in Preview mode

**Actual result:** Next Preview mode opened has the contrast color as was set previously.

**Expected result:** Every new opening of Preview mode has contrast color set to default 'Black on white'.

### DIFF
--- a/src/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
+++ b/src/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
@@ -261,7 +261,9 @@ export default pluginFactory({
      */
     destroy: function destroy() {
         shortcut.remove(`.${this.getName()}`);
-        return this.storage.setItem('itemThemeId', '');
+        return this.getTestRunner().getPluginStore(this.getName()).then(function(itemThemesStore) {
+            return itemThemesStore.setItem('itemThemeId', '');
+        });
     },
 
     /**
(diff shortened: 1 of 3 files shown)


src/plugins/tools/itemThemeSwitcher/itemThemeSwitcher.js
```diff
@@ -260,8 +260,10 @@ export default pluginFactory({
      * Called during the runner's destroy phase
      */
     destroy: function destroy() {
-        this.storage.setItem('itemThemeId', '');
         shortcut.remove(`.${this.getName()}`);
+        return this.getTestRunner().getPluginStore(this.getName()).then(function(itemThemesStore) {
+            return itemThemesStore.setItem('itemThemeId', '');
```
andrey-ponamarev, line 265: Is it possible to use `removeItem` method?

habemuscode: Yes, thanks :)

PR #282, opened 2020-07-14
[TCA-648] WTP-1576 - Speed control is off screen with initial placement

atsymuk:
Related to : https://oat-sa.atlassian.net/browse/TCA-648
 
Make sure that component inside the container after resize or settings expanding
  
#### How to test

- prepare a delivery with APIP content and TTS plugin enabled
- execute the delivery as a test taker
- try to expand TTS component settings and check if it shift to stay on the page
- try to resize the page and check if component moved with page
  
#### Dependencies
   
Companion PR :
 - [ ] https://github.com/oat-sa/<EXT>/pull/<ID>

### DIFF
--- a/src/plugins/tools/apipTextToSpeech/textToSpeech.js
+++ b/src/plugins/tools/apipTextToSpeech/textToSpeech.js
@@ -308,17 +308,12 @@ function maskingComponentFactory(container, config) {
          */
         toggleSettings() {
             const isSettings = this.is('settings');
-            // offset from right
-            const offsetFromRight = 10;
 
             this.setState('settings', !isSettings);
 
             // if settings was enabled make sure that component still inside the container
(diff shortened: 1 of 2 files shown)


src/plugins/tools/apipTextToSpeech/textToSpeech.js
```diff
@@ -308,9 +308,30 @@ function maskingComponentFactory(container, config) {
...
+                this.moveTo(x > maxXPosition ? maxXPosition : x, y);
+            }
         },
+        /**
+         * Handle browser resize
+         */
+        handleResize() {
+            // offset from right
+            const offsetFromRight = 10;
+            const { x, y } = this.getPosition();
```
bziondik, line 330: please, reduce duplication. Maybe move this lines to one function

PR #296, opened 2020-08-04
Feature TBD-46: stop sending csrf tokens to get item request handlers

wazelin:
- [TBD-46](https://oat-sa.atlassian.net/browse/TBD-46) feat: stop sending CSRF tokens to GET item request handlers
- [TBD-46](https://oat-sa.atlassian.net/browse/TBD-46) chore(version): bump a minor one

ℹ️ Will be incorporated by `extension-tao-testqti` [#1891](https://github.com/oat-sa/extension-tao-testqti/pull/1891)

### DIFF
--- a/src/proxy/cache/proxy.js
+++ b/src/proxy/cache/proxy.js
@@ -278,7 +278,7 @@ export default _.defaults(
              *
              * @returns {Promise} resolves with the action result
              */
-            this.requestNetworkThenOffline = function requestNetworkThenOffline(url, action, actionParams, deferred, noToken = false) {
+            this.requestNetworkThenOffline = function requestNetworkThenOffline(url, action, actionParams, deferred, noToken) {
                 var testContext = this.getDataHolder().get('testContext');
                 var communicationConfig = self.configStorage.getCommunicationConfig();
 
@@ -290,7 +290,7 @@ export default _.defaults(
                     } else {
                         //action is not synchronizable
                         //fallback to direct request
-                        request = self.request(url, actionParams, void 0, noToken);
+                        request = self.request(url, actionParams, void 0, noToken || false);
                         request.then(function(result) {
                             if (self.isOffline()) {
                                 return self.scheduleAction(action, actionParams);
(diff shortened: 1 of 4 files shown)


src/proxy/cache/proxy.js
```diff
@@ -286,7 +290,7 @@ export default _.defaults(
                     } else {
                         //action is not synchronizable
                         //fallback to direct request
-                        request = self.request(url, actionParams);
+                        request = self.request(url, actionParams, void 0, noToken || false);
```
Alroniks, line 293: Why just not use simple `undefined` instead of `void 0`. Which reason? I don't think that somebody will change default value of this global variable.

btamas: Because https://eslint.org/docs/rules/no-undefined rule in `eslint` is enabled, so you cannot use it.
There are 2 solutions:
1. `void 0`
2. `let noop; request(url, actionParams, noop, noToken || false)`

habemuscode: Third option can be some `' '` string empty or null?

bziondik: I prefer the second option

andrey-ponamarev: IMHO when function use more than 2 parameters and some of them is optional it's better to put them to `Object` and use it as one parameter.
It brings a better readability when function is called. Another benefit is an easy and a safety to extend new parameters.

```javascript
// It not possible to guess why do we call functionwith "void 0", "true" 🤷‍♂️ ?
request(this.configStorage.getItemActionUrl(itemIdentifier, 'getItem'), params, void 0, true)

// VS

request({
     url: this.configStorage.getItemActionUrl(itemIdentifier, 'getItem'),
     params
})
```

The suggestion is change signature of 
```javascript
this.requestNetworkThenOffline` = function requestNetworkThenOffline(requestParams) {
   const {url, action, actionParams, deferred, noToken} = requestParams
}

wazelin: I totally agree, and that's what I wanted to do at first, but it means introducing a breaking change into the method's signature.
And the method can be used by some of the proxies, extending it.

PR #298, opened 2020-08-06
Fix/BSA-209/move header message from unanswered stat

bziondik:
Related to: https://oat-sa.atlassian.net/browse/BSA-209
https://oat-sa.atlassian.net/browse/BSA-199

Moved header message outside unanswered statistic message.

How to test:

- set `'test-taker-unanswered-items-message' => false,` in `config/taoQtiTest/testRunner.conf.php`
- create a test with an item and set navigation warnings `Display Next Part Warning` or `Display End Test Warning`
- create delivery
- run delivery, don't answer item
- click Next
- check a popup, should not be a message about unanswered items

### DIFF
--- a/src/helpers/messages.js
+++ b/src/helpers/messages.js
@@ -57,6 +57,8 @@ function getHeader(scope) {
     } else if (scope === 'part') {
         return `<b>${__('You are about to submit this test part.')}</b><br><br>`;
     }
+
+    return '';
 }
 /**
  * Build message if not all items have answers
--- a/src/helpers/messages.js
+++ b/src/helpers/messages.js
@@ -45,8 +45,6 @@ function getExitMessage(scope, runner, message = '', sync) {
 /**
  * Build message if not all items have answers
  * @param {String} scope - scope to consider for calculating the stats
- * @param {Object} runner - testRunner instance
- * @param {Boolean} sync - flag for sync the unanswered stats in exit message and the unanswered stats in the toolbox. Default false
  * @returns {String} Returns the message text
  */
 function getHeader(scope) {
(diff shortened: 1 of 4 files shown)


src/helpers/messages.js
```diff
@@ -40,9 +40,24 @@ function getExitMessage(scope, runner, message = '', sync) {
...
     }
 
-    return `${itemsCountMessage} ${message}`.trim();
+    return `${getHeader(scope)}${itemsCountMessage} ${message}`.trim();
+}
+/**
+ * Build message if not all items have answers
+ * @param {String} scope - scope to consider for calculating the stats
+ * @param {Object} runner - testRunner instance
+ * @param {Boolean} sync - flag for sync the unanswered stats in exit message and the unanswered stats in the toolbox. Default false
```
wazelin, line 49: I don't see those arguments.
```suggestion
```

src/helpers/messages.js
```diff
@@ -40,9 +40,24 @@ function getExitMessage(scope, runner, message = '', sync) {
...
+ * @returns {String} Returns the message text
+ */
+function getHeader(scope) {
+    if (scope === 'section' || scope === 'testSection') {
+        return `<b>${__('You are about to leave this section.')}</b><br><br>`;
+    } else if (scope === 'test' || scope === 'testWithoutInaccessibleItems') {
+        return `<b>${__('You are about to submit the test.')}</b><br><br>`;
+    } else if (scope === 'part') {
+        return `<b>${__('You are about to submit this test part.')}</b><br><br>`;
+    }
```
wazelin, line 59: I realize, this shouldn't event be a case, but let's have a default `return` value, just to be fully compliant with the method's contract.
```suggestion
    }

    return '';
```

PR #301, opened 2020-08-12
Feature TAO-10467 - Better highlight of a current delivery item label in review mode

habemuscode:
**Related to this issue**

It's related to this issue but for product we will implement variable `$info` that is some blue color

https://oat-sa.atlassian.net/browse/TAO-10467

**Description**

**As** test taker
**I would like** when browsing the review screen/navigator to have a more distinct color on the current Item
**so that** that I can identify instantly in which item I am on in respect to the whole section/test part.

**Context**

Ability to mark the color of the page on which you are in the preview on the left side of the page a little more so that it is quicker for candidates to identify it? For example, by displaying it in blue? In this way, the candidate can identify at a glance which page he or she is on in the preview. (Because at the moment there is only a very slight difference in color between the pages viewed and the page on which the candidate is on).

**Acceptance criteria**

- AC1

**GIVEN** test taker is during delivery execution which has enabled the review screen/navigator
**WHEN** test taker browses on review screen/navigator
**THEN** the current item on the review panel menu **MUST** be colorized

**Expected result**

imagen

### DIFF
--- a/scss/inc/_navigator.scss
+++ b/scss/inc/_navigator.scss
@@ -424,6 +424,7 @@ $borderColor: $textHighlight;
         background: white();
         &.active {
             background: $info;
+            color: $textHighlight;
         }
         &:hover {
             background: lighten($lightBlueGrey, 3);
--- a/scss/inc/_navigator.scss
+++ b/scss/inc/_navigator.scss
@@ -422,8 +422,12 @@ $borderColor: $textHighlight;
     }
     .qti-navigator-item {
         background: white();
-        &.active, &:hover {
+        &.active {
             background: $info;
+            color: $textHighlight;
+        }
+        &:hover {
+            background: darken($info, 10);
             color: $textHighlight;
         }
         &.disabled {
--- a/scss/inc/_navigator.scss
+++ b/scss/inc/_navigator.scss
@@ -423,7 +423,7 @@ $borderColor: $textHighlight;
     .qti-navigator-item {
         background: white();
         &.active {
-            background: lighten($lightBlueGrey, 5);
+            background: $info;
         }
         &:hover {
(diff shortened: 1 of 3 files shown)


habemuscode:
For product

imagen

habemuscode:
Have UX approval

scss/inc/_navigator.scss
```diff
@@ -422,11 +422,9 @@ $borderColor: $textHighlight;
     }
     .qti-navigator-item {
         background: white();
-        &.active {
-            background: lighten($lightBlueGrey, 5);
-        }
-        &:hover {
-            background: lighten($lightBlueGrey, 3);
+        &.active, &:hover {
```
wazelin, line 425: Those states had different colors before.
Do we expect to use the same for both?

habemuscode: As I understand yes, because with previous colors on hover, the issue remains, we can't see properly the text. I will double-check

habemuscode: As I talk, customer (and probably TAO) only want to change one, but we have an issue when you hover an active element. So, with Luc and Kristina we are moving this for see what customer wants on hover, and what we want on product.

https://files.slack.com/files-pri/T04V23RQT-F018TEFPZ6X/image.png

jsconan:
Looks good for the active state. Waiting for the remaining fix on the hover.

habemuscode:
After the talk with UX, the solution was changed to the `$info` color on the selected item and a dark version for on hover. This solve 2 issues at the same time:

- We have different colors, for active and hover state, that is better for avoiding customer confusing
- We don't have the issue when select and hover the same item, with light version appears strange

**Final result**

imagen

PR #324, opened 2020-10-30
Fix/con 53/items title

tikhanovichA:
https://oat-sa.atlassian.net/browse/CON-53

We have some cases when a title for an item should be empty. Because of that, we don't need to generate the title for a current item for example.


andrey-ponamarev (approved):
- [ ] The feature is working correctly on my local machine (if applicable)

PR #337, opened 2020-11-18
Feature/TR-121/Add a plugin to pause on error

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TR-121

Add a plugin that displays a message upon error. It then allows either to pause the test or to reload the page.

How to test:
- run the unit test: `npm run test`
- install the plugin in TAO, pass a test and trigger an error (example: you may throw an exception in the getItem action)

### DIFF
--- a/src/plugins/controls/connectivity/pauseOnError.js
+++ b/src/plugins/controls/connectivity/pauseOnError.js
@@ -1,0 +1,65 @@
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2020 (original work) Open Assessment Technologies SA ;
+ */
+
+import __ from 'i18n';
+import pluginFactory from 'taoTests/runner/plugin';
+import dialogTpl from 'taoQtiTest/runner/plugins/controls/connectivity/pauseOnError.tpl';
+
+const name = 'pauseOnError';
+const dialogMessage = {
+    title: __('Something unexpected happened.'),
+    message: __('Please try reloading the page or pause the test. If you pause, you will be able to resume the test from this page.')
+};
+const dialogConfig = {
+    buttons: {
+        labels: {
+            ok: __('Pause the test'),
+            cancel: __('Reload the page')
+        }
+    }
+};
+const pauseMessage = __('You encountered an unexpected issue. The test has been suspended.');
+const pauseContext = {
+    reasons: {
+        category: 'technical',
+        subCategory: 'error'
+    },
+    message: pauseMessage,
+    originalMessage: 'Due to an unexpected issue the test has been suspended.'
+};
+
+export default pluginFactory({
+    name,
+
+    /**
+     * Initialize the plugin (called during runner's init)
+     */
+    init() {
+        const testRunner = this.getTestRunner();
+        const returnToHome = () => testRunner.trigger('pause', pauseContext);
+        const reloadPage = () => testRunner.trigger('reloadpage');
+        const processError = () => {
+            testRunner
+                .on('reloadpage', () => window.location.reload())
+                .trigger('disablenav disabletools hidenav')
+                .trigger(`confirm.${name}`, dialogTpl(dialogMessage), returnToHome, reloadPage, dialogConfig);
+        };
+
+        testRunner.on('error', processError);
+    }
+});
--- a/src/plugins/controls/connectivity/pauseOnError.js
+++ b/src/plugins/controls/connectivity/pauseOnError.js
@@ -33,13 +33,11 @@ const dialogConfig = {
         }
     }
 };
-const pauseMessage = __('You encountered an unexpected issue. The test has been suspended.');
 const pauseContext = {
     reasons: {
         category: 'technical',
         subCategory: 'error'
     },
-    message: pauseMessage,
     originalMessage: 'Due to an unexpected issue the test has been suspended.'
 };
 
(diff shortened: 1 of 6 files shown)


src/plugins/controls/connectivity/pauseOnError.js
```diff
@@ -0,0 +1,65 @@
...
+    }
+};
+const pauseMessage = __('You encountered an unexpected issue. The test has been suspended.');
+const pauseContext = {
+    reasons: {
+        category: 'technical',
+        subCategory: 'error'
+    },
+    message: pauseMessage,
+    originalMessage: 'Due to an unexpected issue the test has been suspended.'
```
btamas, line 43: Shouldn't it be translated?

test/plugins/controls/connectivity/pauseOnError/test.js
```diff
@@ -0,0 +1,173 @@
...
+                    })
+                    .init();
+            }))
+            .catch(err => {
+                assert.pushResult({
+                    result: false,
+                    message: err
+                });
+            })
+            .then(done);
```
btamas, line 121: Sorry for the long rewrite, but I do not understand the purpose of using Promise. Isn't my solution is the same?

- Assertion count check could check that every assert was run. No setTimeout necessary.
- Why do you return with Promise.reject in event handler? `false` is enough, isn't it?

```suggestion
                assert.expect(6);
                runner
                    .on('init', () => {
                        assert.ok(true, 'Runner has been initialized');
                        runner.trigger('error');
                    })
                    .on('confirm.*', (message, accept, cancel, options) => {
                        assert.equal(typeof message, 'string', 'String message provided');
                        assert.equal(typeof accept, 'function', 'Accept callback provided');
                        assert.equal(typeof cancel, 'function', 'Cancel callback provided');
                        assert.equal(typeof options, 'object', 'Options provided');
                        cancel();
                    })
                    .before('pause', () => {
                        assert.ok(false, 'The pause should not be triggered!');
                        runner.destroy();
                    })
                    .before('reloadpage', () => {
                        assert.ok(true, 'A page reload has been triggered');
                        runner.destroy();
                        return false;
                    })
                    .on('destroy', done)
                    .init();
```

test/plugins/controls/connectivity/pauseOnError/test.js
```diff
@@ -0,0 +1,173 @@
...
+
+
+    });
+
+    QUnit.test('error then pause', assert => {
```
btamas, line 126: I think my proposed concept can be used here as well.

wazelin:
Something's fishy about that test run.
```
{ Error: Test run could not finish in 30000ms

    at runQunitPuppeteer.then.catch.e (/srv/data/jenkins/workspace/sa_tao-test-runner-qti-fe_PR-337/node_modules/@oat-sa/tao-qunit-testrunner/src/testrunner.js:96:27)

  path: 'test/plugins/content/keyNavigation/keyNavigation/test.html' }
```

src/plugins/controls/connectivity/pauseOnError.js
jsconan: Not this one. This is a text that will be logged.

test/plugins/controls/connectivity/pauseOnError/test.js
jsconan: >  Assertion count check could check that every assert was run. No setTimeout necessary.
Indeed, but in this case, the unit test will hang if the tested code does not perform as expected. It is better to control the flow.

>  Why do you return with Promise.reject in event handler? false is enough, isn't it?
`return false` is the legacy way. A compatibility layer still allows it, but the proper way is to use a `Promise.reject`.

jsconan:
> Something's fishy about that test run.
> 
> ```
> { Error: Test run could not finish in 30000ms
> 
>     at runQunitPuppeteer.then.catch.e (/srv/data/jenkins/workspace/sa_tao-test-runner-qti-fe_PR-337/node_modules/@oat-sa/tao-qunit-testrunner/src/testrunner.js:96:27)
> 
>   path: 'test/plugins/content/keyNavigation/keyNavigation/test.html' }
> ```

Yes, there is a lot of tests to run, and some are very long and complex, like for the `keyNavigation`. The default timeout of 30s may not be enough. Unfortunately, it is hardcoded in the [QUnit test runner](https://github.com/oat-sa/tao-qunit-testrunner-fe/blob/311bc6b1ecfe4c86166e6da5a19698d911444673/src/testrunner.js#L78), and it doesn't look like it can be superseded.

test/plugins/controls/connectivity/pauseOnError/test.js
btamas: > Assertion count check could check that every assert was run. No setTimeout necessary.
> Indeed, but in this case, the unit test will hang if the tested code does not perform as expected. It is better to control the flow.

I do not agree. It makes the test more complicated for no reason. Most of the time the test runs successfully, so there is no hang at all. But if you would like to avoid long hanging, then set a shorter timeout for the test.

btamas:
> > Something's fishy about that test run.
> > ```
> > { Error: Test run could not finish in 30000ms
> > 
> >     at runQunitPuppeteer.then.catch.e (/srv/data/jenkins/workspace/sa_tao-test-runner-qti-fe_PR-337/node_modules/@oat-sa/tao-qunit-testrunner/src/testrunner.js:96:27)
> > 
> >   path: 'test/plugins/content/keyNavigation/keyNavigation/test.html' }
> > ```
> 
> Yes, there is a lot of tests to run, and some are very long and complex, like for the `keyNavigation`. The default timeout of 30s may not be enough. Unfortunately, it is hardcoded in the [QUnit test runner](https://github.com/oat-sa/tao-qunit-testrunner-fe/blob/311bc6b1ecfe4c86166e6da5a19698d911444673/src/testrunner.js#L78), and it doesn't look like it can be superseded.

I think that is a global timeout, but locales can be set

test/plugins/controls/connectivity/pauseOnError/test.js
jsconan: Well, controlling the flow of a unit test allows making sure the expected behavior is performing properly, and there is no unknown. This is also especially useful in the context of TDD.
But anyway, I agree this construction is not useful for this particular case. At least removing it doesn't harm.

btamas:
> Something's fishy about that test run.
>   path: 'test/plugins/content/keyNavigation/keyNavigation/test.html' }
> ```

Maybe 30sec is not enough for the test to finish. It can happen if your computer is not fast enough to run 5 parallel test in the same time. You can decrease the parallel test count with `PARALLEL_TESTS` environment variable like `PARALLEL_TESTS=2 npm test`

btamas:
- [ ] New code is respecting code style rules
- [ ] New code is respecting best practices
- [ ] New code is not subject to concurrency issues (if applicable)
- [ ] The feature is working correctly on my local machine (if applicable)
- [ ] Acceptance criteria are respected
- [ ] Pull request's target is not `master`

btamas (approved):

wazelin (approved):

meefox (approved):

PR #352, opened 2021-02-10
fix/TR-485-popup-message-fix-french-dutch-german

lecosson:
Related to https://oat-sa.atlassian.net/browse/TR-485

- shrink test navigation warning popup messages for dutch, french and german. This PR removes exceed dot in popups messages.

Required:
 - https://github.com/oat-sa/extension-tao-bosa-selor/pull/84
 - https://github.com/oat-sa/extension-tao-testqti/pull/1969

### DIFF
--- a/src/helpers/messages.js
+++ b/src/helpers/messages.js
@@ -38,7 +38,7 @@ function getExitMessage(scope, runner, message = '', sync, submitButtonLabel) {
     if (messageEnabled) {
         itemsCountMessage = getUnansweredItemsWarning(scope, runner, sync);
 
-        if (itemsCountMessage) {
+        if (itemsCountMessage.length > 1) {
             itemsCountMessage += '.';
         }
     }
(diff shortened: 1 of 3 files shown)


src/helpers/messages.js
```diff
@@ -70,11 +70,16 @@ function getHeader(scope) {
...
  */
 function getActionMessage(scope, submitButtonLabel = __('OK')) {
+    var msg;
     switch (scope) {
         case 'section':
         case 'testSection':
         case 'part':
-            return `${__('Click "%s" to continue', submitButtonLabel)}.`;
+            msg = __('Click "%s" to continue', submitButtonLabel);
+            if (msg.length > 1) {
```
btamas, line 79: This condition is always true, because you define `msg` one line before and it has a value.

jsconan, line 79: Indeed, this condition won't work as expected.
Moreover, it will partially address the issue.
I'd suggest adding a `trim()`

```suggestion
            msg = __('Click "%s" to continue', submitButtonLabel).trim();
            if (msg) {
```

src/helpers/messages.js
```diff
@@ -38,7 +38,7 @@ function getExitMessage(scope, runner, message = '', sync, submitButtonLabel) {
     if (messageEnabled) {
         itemsCountMessage = getUnansweredItemsWarning(scope, runner, sync);
 
-        if (itemsCountMessage) {
+        if (itemsCountMessage.length > 1) {
```
jsconan, line 41: This may not work as expected.

I'd suggest keeping the previous code, but adding a `trim()`

jsconan (changes requested):
Unfortunately, the fix is not resilient enough. It will fail if the blank translation is more than 1 char.

src/helpers/messages.js
lecosson: Not always, if argument translated to empty string

btamas: You are right. My bad.

btamas (approved):

jsconan (approved):
**Review Checklist**

PR #398, opened 2021-06-07
tr-888/right-to-left-review-panel

marpesia:
Related to [TR-888](https://oat-sa.atlassian.net/browse/TR-888)

To test it:
- Run as guest with according locale in generis.conf.php: define('DEFAULT_ANONYMOUS_INTERFACE_LANG','ar-arb'); at line 47

### DIFF
--- a/scss/inc/_rtl.scss
+++ b/scss/inc/_rtl.scss
@@ -6,6 +6,7 @@ body {
   &.delivery-scope {
     &[dir=rtl] {
       @import 'rtl/title-bar';
+      @import 'rtl/review-panel';
     }
   }
 }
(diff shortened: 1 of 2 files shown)


btamas (changes requested):
- The name of the branch is not conventional. It should start with fix, feature etc, not with ticket number.
- Description should contain what did you modified. "Implement RTL mode of blablabla"
- I have problem with design as well. If you check my screenshot, some part is misaligned:
  - the up/down arrow that opens a section
  - the stat numbers (1/17 etc)
  - stat number near Overview

Screen Shot 2021-06-08 at 08 00 46

PR #440, opened 2021-11-03
feature: TR-2486. Add new layout for review panel

svoronuk:
Related to: https://oat-sa.atlassian.net/browse/TR-2486

New design: https://projects.invisionapp.com/share/V411RRJ8TRJM#/screens/458873303

Missing info from task:
- we don't have final design with dimensions/styles etc, so ui team suggest me to do as close as possible
- each item in panel should have states: Answered, Unanswered, Disabled Answered/Disabled Unanswered and Bookmarked. Be careful "Disabled Answered/Disabled Unanswered" states are from liner behavior which is not implemented
- keyboard/accessibility/zooming/resizing/resolution should work as before (or close?). No schemas/designs/...

Main ideas: 
1. i created new templates and return some old classes, which is important for restoring  behavior. For example, for restoration keyboard behavior i have to return empty `filters` block, just because it is anchor for keyNavigation handler. 
2. In scss file i had a lot `!important` directives for overwriting default tao styles which is not compatible with new UI + i do not group classes a lot, to decrease classes nesting.

If you want to restore old behavior - remove property from config: https://github.com/oat-sa/extension-tao-testqti/blob/5d7592e294af7275959c53a2397127e5641f17dd/config/default/testRunner.conf.php#L396

TODO:

- [x] AC3  
- [x] By unknown reason current indicator and focusing border are miscentered on 1-2px...
- [x] add classes in _navigatorFizzy.scss to nullify styles from _navigator.scss
- [x] unit tests
- [x] release and merge https://github.com/oat-sa/tao-core-ui-fe/pull/386
- [ ] ~~new behavior ?~~


scss/inc/_navigatorFizzy.scss
```diff
@@ -0,0 +1,312 @@
...
+        &.active .indicator {
+            position: absolute;
+            display: block;
+            z-index: 1;
+            color: $colorTextDefault;
+            height: 1.4rem;
+            min-width: 1.6rem;
+            background-color: $backgroundColor;
+            top: unset;
+            bottom: -1.6rem;
```
oatymart, line 146: The value seems a little off (it's clipping on the button). With -1.7rem or less, it's ok.
Screenshot 2021-11-05 at 09 33 37

scss/inc/_navigatorFizzy.scss
```diff
@@ -0,0 +1,312 @@
...
+        /* reset parent style */
+        padding: 0 !important; /* instead of rewriting 6-classes selector on focusing */
+        border: none !important; /* instead of rewriting 6-classes selector on focusing */
+
+        &.disabled {
+            background-color: unset !important;
+            color: $colorTextDisabled !important;
+            border-color: $colorTextDisabled !important;
+            opacity: 1 !important;
+            text-shadow: none !important;
```
oatymart, line 101: Would be great to avoid all these `!important`s. I just tried removing them all, and the styles are still applied. (It changes `.disabled:hover` styling, but we could fix that using the selector only.)

If there's some case I didn't see yet, you can explain and we can discuss it.

oatymart: Sorry, I just realised I'm talking about the `.disabled .step` styles. But the point is the same - we should try to find some way to not use `!important` at all.

scss/inc/_navigatorFizzy.scss
```diff
@@ -0,0 +1,312 @@
...
+            &:hover {
+                color: $colorTextDefault;
+                text-decoration: none;
+                cursor: pointer;
+            }
+        }
+    }
+
+    /* tree */
+    .qti-navigator-tree {
```
oatymart, line 75: This tree element gets a scrollbar when its content grows (I think when the button on the final line gets an indicator under it). Since the `<aside>` is also scrollable, this caused - in my case - a double scrollbar and a wrapping of the buttons (4 per row became 3 per row). I'm not exactly sure why the `position:absolute` indicator element does this. Maybe we can avoid the situation by leaving extra space for it?
Screenshot 2021-11-05 at 10 01 03

src/plugins/navigation/review/navigator.js
```diff
@@ -552,6 +559,14 @@ function navigatorFactory(config, map, context) {
         navigator.trigger('jump', position);
     }
 
+    if (config["reviewLayout"] === 'fizzy') {
```
oatymart, line 562: Are you sure the default layout is still working ok? I changed my value in `config/taoQtiTest/testRunner.conf.php` to `'not fizzy'`, and I get this strange rendering:
Screenshot 2021-11-05 at 10 15 03

scss/inc/_navigatorFizzy.scss
```diff
@@ -0,0 +1,312 @@
...
+$backgroundColor: #f2f2f2;
+$radiusCircular: 50%;
+$sizeDefault: 4rem;
+$hitboxSize: 6.25rem;
+$borderThin: 0.125rem;
+$colorTextDisabled: hsl(0, 0%, 45%);
+$colorBgDefault: white;
+$borderMedium: 0.25rem;
+$borderMediumPlus: 0.375rem;
+$colorTextDefault: hsl(0, 0%, 12%);
```
oatymart, line 14: Value seems very similar to `$textColor` from `tao-core-ui/scss/inc/_colors.scss` - can we re-use it here?

scss/inc/_navigatorFizzy.scss
```diff
@@ -0,0 +1,312 @@
...
+$colorBgInverted: $colorTextDefault;
+$colorTextInverted: white;
+$colorPrimaryHighlight: hsl(208, 100%, 32%);
+$colorSuccess: hsl(164, 54%, 25%);
+$colorWarning: hsl(356, 65%, 40%);
```
oatymart, line 19: Do we really have green and red in this UI?

oatymart (changes requested):
So far, the implementation of the new design looks quite good. Take care about the scrollbar issue, and the `!important` CSS - if there's any way to avoid it, we should.

Regarding architecture, I don't have an idea if there is a proper way to override the templates. It looks ok to me, but @jsconan has more expertise.

scss/inc/_navigatorFizzy.scss
svoronuk: Updated

scss/inc/_navigatorFizzy.scss
svoronuk: I removed `!important` where it is possible. 

Fight with `!important` is not so easy) We can wrap all files with basic styles in `tao-core-ui-fe`, like:
```
:not[.cg]{
   /* current global styles */
  @import "inc/bootstrap";
  @import "inc/fonts";
  @import "inc/base";
}

//in navigationFizzy.tpl
<div class="cg qti-panel qti-navigator fizzy">
```
But it is much more work than task required. It may be a part of PoC for implementing LDS in CG.
In terms of this task i do not see easy solution.

scss/inc/_navigatorFizzy.scss
svoronuk: I fixed double scroll, but i a little another way. I didn't catch exactly this issue, but you can check.

scss/inc/_navigatorFizzy.scss
svoronuk: No) Missed part of LDS styles

scss/inc/_navigatorFizzy.scss
svoronuk: I didn't use general colors, because this component (by design team view ) should be the same as in LDS. If i change colors a little now - we later will change it back to this colors, when starting injecting LDS into CG. I prefer to stay colors as is.

src/plugins/navigation/review/navigator.js
svoronuk: Yeah. Here my mistake. When i developed styles in separate file i import it after the `_navigator.scss`, so when layout switch back, it has rewritten styles. I'll change order and adopt new styles for this case.

scss/inc/_navigatorFizzy.scss
oatymart: Fix works.

src/plugins/navigation/review/navigator.js
svoronuk: fixed

src/plugins/content/accessibility/keyNavigation/strategies/itemNavigation.js
```diff
@@ -90,7 +90,10 @@ export default {
          * @param {Object} cursor - The cursor definition supplied by the keyNavigator
          * @returns {jQuery} - The selected choice element
          */
-        const getQtiChoice = cursor => cursor && cursor.navigable.getElement().closest('.qti-choice');
+        const getQtiChoice = function (cursor){
+            console.log(cursor);
```
olga-kulish, line 94: Please remove `console.log`. I also saw stuff like `logger.debug` being used, but didn't check how to configure log level.

src/plugins/navigation/review/navigator.js
```diff
@@ -552,6 +559,14 @@ function navigatorFactory(config, map, context) {
...
+        navigatorTpl = fizzyTpl;
+        navigatorTreeTpl = fizzyTreeTpl;
+        // hack to not allow activate/deactivate bookmarking on icon click in item button
```
olga-kulish, line 565: Just a question (you don't have to change anythning): so ideally you should have passed `isFizzyLayout` condition through to js methods below, but since you don't want to make many changes, you use this workaround. Right?

src/plugins/navigation/review/review.js
```diff
@@ -37,15 +37,15 @@ import navigatorFactory from 'taoQtiTest/runner/plugins/navigation/review/naviga
 var buttonData = {
     setFlag: {
         control: 'set-item-flag',
-        title: __('Flag the current item for later review'),
-        icon: 'anchor',
-        text: __('Flag for Review')
+        title: __('Bookmark the current question for later review'),
```
olga-kulish, line 40: I think wording and icon will change for 'normal' (not fizzy) layout too? Is that ok - shouldn't old layout stay the same?

olga-kulish (commented):
First look (I'll test more next) - looks quite nice!

Question:
1) it's ok that we can't jump to items that were not visited yet, right?
2) it's ok that for linear test we see _'In this part of the test navigation is not allowed_.' message, right?

src/plugins/navigation/review/navigator.js
```diff
@@ -719,4 +743,6 @@ function navigatorFactory(config, map, context) {
     return navigator.init(config);
 }
 
+Handlebars.registerPartial('bubbleButton', bubbleButton);
```
jsconan, line 746: **nitpick:** We should avoid adding partials for nested components. Partials are global, and having a component registering specific partial may introduce conflicts and unpredictable issues without having a proper LDS way of doing it. While this may be a good practice in a small project, it is not the case here due to the complexity.

It would be better to either inline the partial (it is used only once for the time being), or to render the element just in time.

src/plugins/navigation/review/review.js
jsconan: **nitpick:** Caution, this will impact every version of the component. It will introduce unwanted changes to the "standard" mode.

jsconan (changes requested):
Good job on a not so easy task.
I did a code review only so far. I saw several problems that should be solved first.

src/plugins/navigation/review/navigator.js
svoronuk: Yes) All this task about workarounds and compromises.

src/plugins/content/accessibility/keyNavigation/strategies/itemNavigation.js
svoronuk: My bad

scss/inc/_navigatorFizzy.scss
```diff
@@ -0,0 +1,331 @@
+@import 'inc/bootstrap';
+
+$lineHeight: 3rem;
+$borderWidth: 1px;
+$backgroundColor: #f2f2f2;
```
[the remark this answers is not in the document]
svoronuk: My main vision, that such style changes - new layout (NG layout? )) ), so i wanted maximally split current theme/basic styles from new layout and may be we could created new theme, if some task were not removed from sprint. I scare that clean this styles later will be harder, than just remove some variables from exact file. However at this moment i see that this approach is not aligned with approved, so i will do as you suggest and may be later we will change this.

src/plugins/navigation/review/review.js
svoronuk: I used the same flag here also

scss/inc/_navigatorFizzy.scss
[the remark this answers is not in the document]
svoronuk: All colors moved into corresponding file and wrapped in separate block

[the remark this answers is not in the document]
svoronuk: @jsconan what i should do with this changes? Should i ignore all this files or created 10+ PRs or create PR only for update css, not css.map ? 
image

src/plugins/navigation/review/navigator.js
svoronuk: Unfortunately https://handlebarsjs.com/guide/partials.html#inline-partials implemented in version 4 i just moved this template inside `navigatorBubbles.tpl` and remove imports and partials

src/plugins/navigation/review/navigator.js
```diff
@@ -23,9 +23,15 @@ import _ from 'lodash';
 import component from 'ui/component';
 import autoscroll from 'ui/autoscroll';
 import mapHelper from 'taoQtiTest/runner/helpers/map';
-import navigatorTpl from 'taoQtiTest/runner/plugins/navigation/review/navigator.tpl';
-import navigatorTreeTpl from 'taoQtiTest/runner/plugins/navigation/review/navigatorTree';
-
+import Handlebars from 'handlebars';
```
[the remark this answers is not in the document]
svoronuk: Removed

scss/inc/_navigatorFizzy.scss
svoronuk: I refactored all colors to theme variables, as you suggested

scss/inc/_navigatorFizzy.scss
```diff
@@ -0,0 +1,331 @@
...
+$headerHeight: 4.5rem;
+$panelPadding: 19px;
+
+.qti-navigator.fizzy {
+    .qti-navigator-tree.key-navigation-group.focusin {
+        background: none;
+    }
+}
+.qti-navigator.fizzy.qti-panel {
+    &.qti-panel.qti-panel {
```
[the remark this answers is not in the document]
jsconan: What about something like this? `.test-runner-scope .test-sidebar .qti-navigator.fizzy.qti-panel  .key-navigation-group.focusin`
Maybe can we use a shorted selector too.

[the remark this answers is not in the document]
svoronuk: Used initial selector with additon `.fizzy` class

src/plugins/navigation/review/review.js
```diff
@@ -156,9 +172,11 @@ export default pluginFactory({
 
         const testRunnerOptions = testRunner.getOptions();
         const pluginShortcuts = (testRunnerOptions.shortcuts || {})[this.getName()] || {};
-        const navigatorConfig = testRunnerOptions.review || {
+        const pluginConfig = this.getConfig();
+        let navigatorConfig = testRunnerOptions.review || {
             defaultOpen : false
         };
+        navigatorConfig = Object.assign({}, navigatorConfig, pluginConfig)
```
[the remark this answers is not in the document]
svoronuk: fixed.

src/plugins/navigation/review/review.js
```diff
@@ -64,10 +76,14 @@ var buttonData = {
...
  * Gets the definition of the flagItem button related to the context
  * @param {Boolean} flag - the flag status
+ * @param {Object} config - plugin config
  * @returns {Object}
  */
-function getFlagItemButtonData(flag) {
-    const dataType = flag ? 'unsetFlag' : 'setFlag';
+function getFlagItemButtonData(flag, config = null) {
+    let dataType = flag ? 'unsetFlag' : 'setFlag';
+    if (config && config["reviewLayout"] === 'fizzy') {
```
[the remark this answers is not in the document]
svoronuk: You are right. Let be align with previous approach.

src/plugins/navigation/review/review.js
```diff
@@ -352,7 +373,7 @@ export default pluginFactory({
                 if (isPluginAllowed()) {
                     updateButton(
                         self.flagItemButton,
-                        getFlagItemButtonData(isItemFlagged(map, context.itemPosition))
+                        getFlagItemButtonData(isItemFlagged(map, context.itemPosition), self.getConfig())
```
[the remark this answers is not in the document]
svoronuk: I've changed this.

src/plugins/navigation/review/navigatorBubbles.tpl
```diff
@@ -0,0 +1,32 @@
...
+                <ul aria-label="{{label}}" class="qti-navigator-items plain">
+                    {{#each items}}
+                        <li class="qti-navigator-item {{cls}}" data-id="{{id}}" data-position="{{position}}">
+                            <button
+                                    class="qti-navigator-label step"
+                                    {{#if active}}class="current"{{/if}}
+                                    aria-disabled="{{#if viewed}}false{{else}}true{{/if}}"
+                                    {{#if active}}aria-current='location'{{/if}}
+                                    role="link"
+                                    aria-label="{{index}} of {{../stats.total}} {{icon}}">
```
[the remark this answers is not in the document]
svoronuk: Araceli decided change behavior for this template. I've update code.

src/plugins/navigation/review/navigatorBubbles.tpl
```diff
@@ -0,0 +1,32 @@
...
+                                {{#if active}}<span class="{{#if active}}icon-indicator {{/if}}indicator" aria-hidden="true"></span>{{/if}}
+                                <span class="qti-navigator-icon icon-{{icon}}" aria-hidden="true"></span>
+                                <span class="step-label" aria-hidden="true">{{position}}</span>
```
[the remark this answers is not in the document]
svoronuk: Fixed.

src/plugins/navigation/review/navigatorBubbles.tpl
```diff
@@ -0,0 +1,32 @@
...
+    <li class="qti-navigator-part active" data-id="{{id}}">
+        <ul aria-label="{{label}}" class="qti-navigator-sections plain">
+            {{#each sections}}
+            <li class="qti-navigator-section active" data-id="{{id}}">
+                <span class="qti-navigator-label" title="{{label}}">
+                    <span class="qti-navigator-text">{{label}}</span>
+                </span>
+                <ul aria-label="{{label}}" class="qti-navigator-items plain">
+                    {{#each items}}
+                        <li class="qti-navigator-item {{cls}}" data-id="{{id}}" data-position="{{position}}">
```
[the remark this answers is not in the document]
svoronuk: Fixed.

scss/inc/_navigatorFizzy.scss
```diff
@@ -0,0 +1,331 @@
...
+            text-shadow: none;
+        }
+
+        /* remove number from item button if item was marked for review */
+        &.flagged span.step-label {
+            display: none;
+        }
+        /* remove number from item button if item is informational */
+        &.info span.step-label {
+            display: none;
```
olga-kulish, line 218: Double-check: In NextGen we don't increase item index (and total) for informational items. Is it ok that here we don't? I'm fine with it, just confirming. 

Screenshot_17

bziondik (approved):
- [ ]  New code is covered by tests (if applicable)
- [ ]  Tests are running successfully (old and new ones) on my local machine (if applicable)
- [x]  New code is respecting code style rules
- [x]  New code is respecting best practices
- [x]  New code is not subject to concurrency issues (if applicable)
- [ ]  Feature is working correctly on my local machine (if applicable)
- [x]  Acceptance criteria are respected
- [x]  Pull request title and description are meaningful

scss/inc/_navigatorFizzy.scss
svoronuk: Yeah. This is in progress. I hoped i finish it faster)

svoronuk: I've added solution here

test/plugins/navigation/review/navigator/test.js
```diff
@@ -319,6 +325,48 @@ define([
...
+    QUnit.test('visual test', function(assert) {
+        const ready = assert.async();
+        const $container = $('#outside');
+        const config = Object.assign({}, sample.config, { "reviewLayout": "fizzy"});
+
+        assert.expect(1);
+
+        navigatorFactory(config, sample.map, sample.context)
+        .on('update', function() {
+            assert.ok(true);
```
olga-kulish, line 365: Optional suggestion:
Maybe we can add a check for informational item logic here - That `item.numberTest` is skipped for them?

And maybe a check for some other basic things, like there are N buttons (N= item count), but I'm not sure, since test for 'normal' layout also contains a very simple test-case like this.

olga-kulish (approved):
Great job investigating and implementing all of this!

Still need to update `tao-core-ui-fe` version before merging.
And you can make assertions in unit test more detailed if you prefer.

test/plugins/navigation/review/navigator/test.js
svoronuk: I will add test for new features. This exact code - only rendering the panel, for visual check, so such assertion is normal.

svoronuk: Tests are added

olga-kulish: I've rechecked - all is good

PR #454, opened 2021-12-03
Feature/TR-2846/Manage assets TTL for the caching proxy

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TR-2846

Implement a TTL management for the cached item store.
Since the modified modules were still in ES5, I also took the opportunity to convert them to ES6. For more commodity while reviewing, don't hesitate to look at the commits, the ES6 conversion has been made in separate commits.

How to test: please look at the main PR for instructions.

### DIFF
--- a/src/proxy/cache/itemStore.js
+++ b/src/proxy/cache/itemStore.js
@@ -13,7 +13,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2017 Open Assessment Technologies SA
+ * Copyright (c) 2017-2021 Open Assessment Technologies SA
  */
 
 /**
@@ -28,32 +28,30 @@ import itemPreloaderFactory from 'taoQtiTest/runner/proxy/cache/itemPreloader';
 /**
  * The default number of items to store
  */
-var defaultConfig = {
+const defaultConfig = {
     maxSize: 10,
     preload: false
 };
 
 /**
  * Create an item store
- * @param {Object} [options]
- * @param {Number} [options.maxSize = 10] - the store limit
- * @param {Boolean} [options.preload] - do we preload items when storing them
- * @param {String} [options.testId] - the unique identifier of the test instance, required if preload is true
+ * @param {object} [options]
+ * @param {number} [options.maxSize = 10] - the store limit
+ * @param {boolean} [options.preload] - do we preload items when storing them
+ * @param {string} [options.testId] - the unique identifier of the test instance, required if preload is true
  *
  * @returns {itemStore}
  */
 export default function itemStoreFactory(options) {
-    var config = _.defaults(options || {}, defaultConfig);
+    const config = _.defaults(options || {}, defaultConfig);
 
     //in memory storage
-    var getStore = function getStore() {
-        return store('item-cache', store.backends.memory);
-    };
+    const getStore = () => store('item-cache', store.backends.memory);
 
     //maintain an index to resolve existence synchronously
-    var index = [];
+    let index = [];
 
-    var itemPreloader;
+    let itemPreloader;
     if (config.preload) {
         itemPreloader = itemPreloaderFactory(_.pick(config, ['testId']));
     }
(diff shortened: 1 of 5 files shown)


src/proxy/cache/itemStore.js
```diff
@@ -28,32 +28,43 @@ import itemPreloaderFactory from 'taoQtiTest/runner/proxy/cache/itemPreloader';
...
     //in memory storage
-    var getStore = function getStore() {
-        return store('item-cache', store.backends.memory);
+    const getStore = () => store('item-cache', store.backends.memory);
+
+    //maintain an index to resolve existence synchronously:
+    // - the array allows to manage the cache in order, it is useful to remove the first entries
+    // - the map allows to quickly check each cache entry, and store some internal metadata like the timestamp
+    let list = [];
+    const index = new Map();
```
oatymart, line 56: I don't love the idea of managing the Array and the Map in parallel. The scope is luckily quite limited, but it's still a potential trap for someone refactoring in future. Do you think it can be done using only the Map and a comparison function at the moment of removing the earliest item? Or will that be worse for readability and performance?

src/proxy/cache/itemStore.js
```diff
@@ -67,57 +78,64 @@ export default function itemStoreFactory(options) {
...
                         }
                     }
 
                     //do we reach the limit ? then remove one
-                    if (index.length > 1 && index.length > config.maxSize) {
-                        return self.remove(index[0]).then(function(removed) {
-                            return updated && removed;
-                        });
+                    if (list.length > 1 && list.length > config.maxSize) {
+                        return this.remove(list[0]).then(removed => updated && removed);
```
oatymart, line 138: The key will remain in the index (until pruning), is it intended?

src/proxy/cache/proxy.js
```diff
@@ -33,17 +33,29 @@ import offlineErrorHelper from 'taoQtiTest/runner/helpers/offlineErrorHelper';
...
+
+/**
+ * The default TimeToLive for assets resolving.
+ * Each item comes with a baseUrl that may have a TTL bound to it.
+ * Once this TTL is expired, the assets won't be reachable.
+ * For this reason, we need to remove from the cache items having an expired TTL.
+ * @type {number}
+ * @private
+ */
+const defaultItemTTL = 15 * 60;
```
oatymart, line 58: Please comment seconds or milliseconds. It might be obvious here but better to be explicit, since we have a mix of the two in TAO.

src/proxy/cache/proxy.js
```diff
@@ -55,16 +67,22 @@ export default _.defaults(
...
+        install(config) {
             //install the parent proxy
             qtiServiceProxy.install.call(this);
 
+            const getConfigOption = (name, defaultValue) => {
+                if (config && config.options && config.options.itemCaching) {
+                    defaultValue = parseInt(config.options.itemCaching[name], 10) || defaultValue;
+                }
+                return defaultValue;
+            };
```
oatymart, line 81: This doesn't look quite right to me...
- the function has a generic name but only gets numeric values from within `itemCaching` property (quite specific)
- reassignment to `defaultValue` arg when it would make more sense to use another name

If it's intentionally multi-purpose, but limited, maybe explain with a comment.

src/proxy/cache/proxy.js
```diff
@@ -511,53 +519,58 @@ export default _.defaults(
                 }
             };
 
+            // the proxy options are supplied after the 'init' phase, we need to apply them later
+            this.itemStore.setItemTTL(this.getItemTTL());
```
oatymart, line 523: I don't understand why config options wouldn't be available at `install` or `init` (they seem to be passed)? It looks weird to set again the store TTL every time `getItem` is called.

oatymart (changes requested):
Code review only. The principles look good. I'll test it tomorrow.

src/proxy/cache/proxy.js
jsconan: I know, this is weird. I spend some time trying to workaround this, but in fact, the config is given as the result of the `init` action, called by the proxy init. See https://github.com/oat-sa/tao-test-runner-qti-fe/blob/7bfb016066cf838629b6e7703e340037bbaeafd6/src/provider/qti.js#L613

A refactoring would be needed to supply these values at a more appropriate time, but it is beyond this task.

src/proxy/cache/itemStore.js
jsconan: The remove function is taking care of the index. Nothing is removed outside of this function.

oatymart: Sorry, my bad!

src/proxy/cache/itemStore.js
jsconan: Map objects can keep the insertion order. However, there is no way to access them by position. Eventually, we could rely on the iteration protocol to access the first position, but it may have unexpected behaviour since we may have a race condition that will disturb the insertion order (items are added in burst without waiting for the previous to be set).

We could indeed maintain a position value for each item in the map, and then search for the lower one when trimming. It will have a low impact on performance considering the small amount of entries. But it will also have some complexity.

I was also considering implementing an ordered map in a separate module, offering an API to manage it. But I lacked time and I made the choice to go straight to the essential.

src/proxy/cache/proxy.js
jsconan: I add the time unit in the description.

oatymart (approved):
The code is improved enough, I've tested the proxy/store logic in my local environment via logging and it seems correct to me, and I can't reproduce the media issue on the kitchen deployment.

The "sitting on one item for 24 minutes before playing the video" issue still reproduces, but it could be tackled in a next ticket.

- [x] New code is covered by tests (if applicable)

src/proxy/cache/itemStore.js
```diff
@@ -28,32 +28,66 @@ import itemPreloaderFactory from 'taoQtiTest/runner/proxy/cache/itemPreloader';
...
+
+    // maintain an index to resolve existence synchronously
+    const index = new Map();
+    let lastIndexedPosition = 0;
+
+    // check if a key has expired
+    const isExpired = key => {
+        const meta = index.get(key);
+        if (meta) {
+            return config.itemTTL && Date.now() - meta.timestamp >= config.itemTTL;
```
olga-kulish, line 61: In hypothetical case where asset implementation has no TTL and urls are permanent, what `config.itemTTL` do you expect to be used? `0` or very big number?

olga-kulish (approved):
Well done!

PR #456, opened 2021-12-13
fix: TR-3068. Pasue test with correct response

svoronuk:
Related to: https://oat-sa.atlassian.net/browse/TR-3068

Issue: If user try to move on another item(click next/previous button), while "pause" signal do not come to user test, then qti provider send to server empty item response.

Shortly about the root cause steps:
1. User answered some items and go to one of answered
2. Proctor press "pause"
3. Paused state do not sent to user machine, becasue connection between test and proctor is one-way-directional: user's test cosume information from proctor.
4. User press "next/previous" button (important that this happen before next polling request)
5. Fired "move" event
6. Item response stored in store and called "unload" item flow
7. During "unload",  test runner clear qti provider instance and fire data sync
8. During data sync, test runner send request to proctor to consume actual state
9. Proctor returns "paused" state
10. Fired "pause" event
11. Inside event handler we try to get item state using item runner instance, but at this time qti provider is cleared by previous event, so item runner returns fallback  value for state: `{}`
12. Set empty object into params.

How to test: 
1. Switch repo of tao-item-runner-fe to https://github.com/oat-sa/tao-item-runner-fe/pull/49
2. Switch  repo of tao-item-runner-qti-fe to https://github.com/oat-sa/tao-item-runner-qti-fe/pull/239
3. Follow steps from ticket

_You can also use proctor via UI for testing this._

**Important moments**: 
1. This flow is still have issue if user press "skip" right after pause. This problem have another root cause + BOSA is the only client who uses proctoring + BOSA test never allow "skip", so i notify QA about edge case and did not fix it.
2. My changes also can be simplified just do not sending any item response during "pause" handling, but i left this way as it is more stable and predictable.
3. I didn't add test for this repo, because current test for this file is cover very basic functionality  and didn't cover any events.

### DIFF
--- a/src/provider/qti.js
+++ b/src/provider/qti.js
@@ -492,18 +492,40 @@ var qtiProvider = {
             })
             .on('pause', function(data) {
                 const testContext = self.getTestContext();
-
+                let baseParams = {
+                    itemDefinition: testContext.itemIdentifier,
+                    reason: {
+                        reasons: data && data.reasons,
+                        comment: data && (data.originalMessage || data.message)
+                    }
+                }
                 this.setState('closedOrSuspended', true);
 
-                this.getProxy()
-                    .callTestAction('pause', {
-                        itemDefinition: testContext.itemIdentifier,
-                        itemState: self.itemRunner.getState(),
-                        reason: {
-                            reasons: data && data.reasons,
-                            comment: data && (data.originalMessage || data.message)
+                // get item response depends on availability item runner
+                const itemState = new Promise((resolve, reject) => {
+                    if(self.itemRunner.isCleared()){
+                        const proxy = this.getProxy();
+                        const store = proxy && proxy.itemStore;
+
+                        if(store) {
+                            store.get(testContext.itemIdentifier)
+                            .then(result => {
+                                resolve(result.itemState);
+                            })
+                            .catch(err => reject(err));
+                        } else {
+                            reject(__('Response can not be stored.'));
                         }
-                    })
+                    } else {
+                        resolve(self.itemRunner.getState());
+                    }
+                });
+
+                itemState.then(data =>{
+                    const params = Object.assign({}, baseParams, {itemState: data});
+
+                    self.getProxy()
+                    .callTestAction('pause', params)
                     .then(function() {
                         self.trigger('leave', {
                             code: states.testSession.suspended,
@@ -513,6 +535,7 @@ var qtiProvider = {
                     .catch(function(err) {
                         self.trigger('error', err);
                     });
+                });
             })
             .on('move skip exit timeout pause', function() {
                 stopwatch.stop();


src/provider/qti.js
```diff
@@ -492,18 +492,40 @@ var qtiProvider = {
...
+                            reject(__('Response can not be stored.'));
                         }
-                    })
+                    } else {
+                        resolve(self.itemRunner.getState());
+                    }
+                });
+
+                itemState.then(data =>{
+                    const params = Object.assign({}, baseParams, {itemState: data});
```
olga-kulish, line 525: I have doubts about this approach in general (sorry, don't have enough knowledge about this):

- Instead of getting state from `proxy.itemStore` (are there any guarantees on what and in which state is inside?), could we pass `itemState: null` to `pause` request and ensure backend will keep value already submitted to server in this case?

- it feels strange to rely on itemRunner (`itemRunner.isCleared()`) to detect this, feels like it should be some test-runner state instead, but I don't see anything like test-runner state here.... https://github.com/oat-sa/tao-test-runner-qti-fe/blob/develop/src/provider/qti.js#L234 `context.itemSessionState` but what value does it have here?

- Is it guaranteed that `pause` action (initiated by system) will happen _only immediately after_ `move` action (initiated by user). Because if not, and it can happen in the middle of user answering, we will probably submit answer he didn't finish - is it fine? Shouldn't unfinished answer be saved only if there's `answerCachePlugin`? Though I see passing`itemState` on `pause` was in code before (`itemState: self.itemRunner.getState()` on line 501)

src/provider/qti.js
```diff
@@ -492,18 +492,40 @@ var qtiProvider = {
...
-                            reasons: data && data.reasons,
-                            comment: data && (data.originalMessage || data.message)
+                // get item response depends on availability item runner
+                const itemState = new Promise((resolve, reject) => {
+                    if(self.itemRunner.isCleared()){
+                        const proxy = this.getProxy();
+                        const store = proxy && proxy.itemStore;
+
+                        if(store) {
+                            store.get(testContext.itemIdentifier)
```
olga-kulish, line 511: Is it guaranteed that store.get will return something?

src/provider/qti.js
```diff
@@ -492,18 +492,40 @@ var qtiProvider = {
...
-                        itemDefinition: testContext.itemIdentifier,
-                        itemState: self.itemRunner.getState(),
-                        reason: {
-                            reasons: data && data.reasons,
-                            comment: data && (data.originalMessage || data.message)
+                // get item response depends on availability item runner
+                const itemState = new Promise((resolve, reject) => {
+                    if(self.itemRunner.isCleared()){
+                        const proxy = this.getProxy();
+                        const store = proxy && proxy.itemStore;
```
olga-kulish, line 508: Do all proxies that can be used with this provider have itemStore - is it some required interface? Since you reject if it doesn't exist

jsconan: Good catch Olga!
Indeed, this API is not standard, and the item store is only available in the context of the pre-caching proxy.

I don't think this should be made this way. The test runner implementation must not tamper with the proxy internals.

src/provider/qti.js
jsconan: Totally agree Olga. Very good catch and analysis.

Despite the proposed solution looks to fix the issue, it does not suit well. Actually, it only takes care of one case, which is only occurring when we have both the proctoring and the pre-caching activated.

Moreover, it assumes too much on the proxy internals, which is not safe.

The pause event can occur at any time. The state should be supplied when pausing the test to prevent the user from losing any data. For this particular point, I don't think it is an issue if the user input is not complete. You can consider it as a backup from an interruption. When resuming, the user expects to recover from the left state.

In fact, the solution needs to take care of all cases, but without dealing with consumed components internals. The test needs to maintains and observe the states, and then react accordingly, but without looking up inside the guts of the components.

jsconan (changes requested):
While the issue looks to be fixed, roughly, it only applies in a particular context (proctoring and pre-caching).

Sorry for this, but I think the solution needs to be done in a different way, without tampering with the internals. Please look at the comments made by Olga and I.

PR #457, opened 2021-12-15
fix: TR-3163. Migrate to github actions

svoronuk:
- upgrade some dev tools 
 - migrate to GitHub Actions the following checks in CI:
   - conventional commits
   - unit test
   - coverage
   - eslint 

TODO:
- [x] fix tests. After approving and merging the PR https://github.com/oat-sa/tao-core-ui-fe/pull/406 continuous integration will be restored
- [ ] update `tao-core-ui-fe` dependency

Eslint report job fails because the repo contains too many lint errors, it's not blocking for PRs but highlights the changes required.


package.json
```diff
@@ -62,8 +68,6 @@
         "decimal.js": "10.1.1",
         "dompurify": "1.0.11",
         "eslint": "^7.32.0",
-        "eslint-plugin-es": "^4.1.0",
-        "eslint-plugin-jsdoc": "^36.1.0",
```
jsconan, line 66: **question:** Why is it removed? This is still in use and it does not look to be installed by peers. Removing this might add back IDE complains and will prevent it from working properly...

jsconan (approved):
It looks mostly ok.
Just one remark, non blocking.

package.json
svoronuk: It is dependency from https://github.com/oat-sa/eslint-config-tao/blob/master/package.json

oatymart (approved):
It looks good. I would not bother to wait for the TR-3292 fix, as we can reasonably assume that the test run will complete as expected in future, because it is using the same action as in our other repos.

PR #458, opened 2021-12-16
fix: TR-3068. Restore item state after pause

svoronuk:
Related to: https://oat-sa.atlassian.net/browse/TR-3068

Continue [discussion](https://github.com/oat-sa/tao-test-runner-qti-fe/pull/456) here:

I knew this weaknesses for my solution, but follow me step by step:
1. In time of handling pause we have two situations: or user pressed navigation button before or didn't. If he didn't press anything my code follow current flow and send correct item state to server. If button is pressed, this means we are finished unloading element and we go to point 2
2. We have 3 ways to get item state: a) using caching proxy (our case), b) using local storage (as i understand it is `answerCache` plugin), c) request to server. Returning to our situation, we are in a moment of handling "paused" state (navigation button is already pressed), where item already is unloaded. To get info about the item state we can use three ways i mentioned: a) - i used it, b) - was not available, c) - the worst case. 
3. I absolutely understand that using proxy cache for getting data is risky, as soon as believe in any plugin which can be disabled, so i added various checks which allow/try to get info from a) or b). If all failed, it means we did not have any possibility to restore state ( c) - is not a variant at all) in terms of ticket's ACs. 
4. Let's imagine now, that we have client which uses proctoring, but without proxy cache and caching plugins, what will happen? Ticket's issue is happened and no way to avoid this with or without my fix. Should we take care about this potential case? From my point yes, but we should decide than what we want: change our expectations or add new feature. In any case it is far from ticket scope.
5. `isCleared` method is really bad. I understand this now, by spending couple of hours to deep dive in architecture from calling controller and rendering item. 

All this text above is just a lyric where i focused that didn't have a lot variants to fix issue in terms of ACs and limited time for unblocking regression. 

Current solution i mentioned in previous PR and it has some unclear moments for me on BE side. The core moment - i do not sent the `itemState` during "paused" state. From my experiments old cache replaced by new one field by field. It means, if field is absent it won't be touched. BE still can not confirm this. Based on this assumption fix is working, but i don't like this also, because we depends on how BE handle our "partial" item state.

I think as solution we can change nothing and enable `answerCache` plugin, but need to test this + does customer want to enable something?

Or we can say, that this is not a bug, it is a feature)))

### DIFF
--- a/src/provider/qti.js
+++ b/src/provider/qti.js
@@ -495,15 +495,21 @@ var qtiProvider = {
 
                 this.setState('closedOrSuspended', true);
 
+                let params = {
+                    itemDefinition: testContext.itemIdentifier,
+                    reason: {
+                        reasons: data && data.reasons,
+                        comment: data && (data.originalMessage || data.message)
+                    }
+                };
+
+                const itemState = self.itemRunner.getState();
+                if (Object.keys(itemState).length) {
+                    params = Object.assign({}, params, { itemState });
+                }
+
                 this.getProxy()
-                    .callTestAction('pause', {
-                        itemDefinition: testContext.itemIdentifier,
-                        itemState: self.itemRunner.getState(),
-                        reason: {
-                            reasons: data && data.reasons,
-                            comment: data && (data.originalMessage || data.message)
-                        }
-                    })
+                    .callTestAction('pause', params)
                     .then(function() {
                         self.trigger('leave', {
                             code: states.testSession.suspended,
--- a/src/provider/qti.js
+++ b/src/provider/qti.js
@@ -495,7 +495,7 @@ var qtiProvider = {
 
                 this.setState('closedOrSuspended', true);
 
-                let params = {
+                const params = {
                     itemDefinition: testContext.itemIdentifier,
                     reason: {
                         reasons: data && data.reasons,
--- a/src/provider/qti.js
+++ b/src/provider/qti.js
@@ -505,7 +505,7 @@ var qtiProvider = {
 
                 const itemState = self.itemRunner.getState();
                 if (Object.keys(itemState).length) {
-                    params = Object.assign({}, params, { itemState });
+                    params.itemState = itemState;
                 }
 
                 this.getProxy()


src/provider/qti.js
```diff
@@ -495,15 +495,21 @@ var qtiProvider = {
...
+                let params = {
+                    itemDefinition: testContext.itemIdentifier,
+                    reason: {
+                        reasons: data && data.reasons,
+                        comment: data && (data.originalMessage || data.message)
+                    }
+                };
+
+                const itemState = self.itemRunner.getState();
+                if (Object.keys(itemState).length) {
```
olga-kulish, line 507: It's good that now we don't send `itemState` if we don't know it, but this condition still doesn't look better than `isCleared` - how does test-runner know it's '_null'_ state, not '_user removed all answers state_'? Also in reality it should work, so we can leave it, maybe at least move it to some helper function.
I don't have a suggestion on how to improve it if `context.state` and `context.itemSessionState` are still `interacting` and `unloadItem` method doesn't set any state.

src/provider/qti.js
```diff
@@ -495,15 +495,21 @@ var qtiProvider = {
...
+                    params = Object.assign({}, params, { itemState });
```
olga-kulish, line 508: Now I guess we'll lose answers user changed this time, since `move` failed and his answer wasn't actually submitted. (You've written about `answerCache` in description). It could be ok (we did fix the main bug!), but if not, could we think of this as a continuation of failed `move` action and pass it `itemState` which was supposed to go into `move`? 

`pause` is called [at least] here
a. https://github.com/oat-sa/tao-test-runner-qti-fe/blob/develop/src/provider/qti.js#L714
b. https://github.com/oat-sa/tao-test-runner-qti-fe/blob/develop/src/provider/qti.js#L263 --> our case?
Then for 'b.' something like:
```
function computeNext(action, params, loadPromise) {
   self.trigger('pause', {message : err.message, moveParams: params});
}
...
.on('pause', function(data) {
	this.getProxy().callTestAction('pause', {
		itemState: moveParams && moveParams.itemState && moveParams.itemIdentifier === testContext.itemIdentifier ? moveParams.itemState : self.itemRunner.getState()
```

Note 1: I hope it's fine to have `itemState` without `itemResponses`...
Note 2: If we decide not to send user's latest changes in this case, could it be better not to send `itemState` at all? In case 'a.' itemRunner can be in corrupted state? Though maybe 'pause' is called from other places too and they need state to be preserved.

src/provider/qti.js
olga-kulish: > Though, in reality it should work, so we can leave it. Maybe move it to some helper function.

Maybe here https://github.com/oat-sa/tao-test-runner-qti-fe/blob/master/src/helpers/currentItem.js  but I don't really know

src/provider/qti.js
svoronuk: I think we can add to case _b._  extra params, because it also will be called after clearing item runner and we will get the same situation: item state is cleared. My fix will not work, if "move" action is finished with error.

However we can not avoid sending `itemState` in "pause" handler. For case where user working with test item and in this moment he get "paused" state, storing current state is absolutely necessary, because after resuming we always get actual state from server and user will be very sad, if he loses his temporary state (especially if test - writing an essay).

oatymart (commented):
The approach looks good. I've tested it and my unsubmitted response was restored after proctor pause/resume. I didn't try to explore other cases than the reported bug.

Like it's mentioned in ticket, `answerCache` plugin also solves this issue. But this fix is useful in case some customer removes or doesn't install that plugin.

It would be nice to have test coverage for the case, but I hardly see any tests regarding these parts of the code. Unless you already found something, it will probably be time-consuming to write one.

src/provider/qti.js
olga-kulish: Update: for me `move` always completes with error in this case. But for you (and looks like Martin too) it succeeds. Then this discussion isn't relevant to the main case, so let's forget it.

Screenshot_38

jsconan (approved):
Good job!

I've checked it in various situations, and it looks to work fine. I've tested with the pre-caching enabled and with a vanilla TAO.

Just one minor remark about the code style.

**Review Checklist**
- [ ] New code is respecting code style rules
- [ ] New code is respecting best practices
- [x] Commits are atomic

olga-kulish (approved):
Good job on a difficult task!

oatymart (approved):

PR #469, opened 2022-02-22
Force Safari iOS to reconsider the scrolling area

krampstudio:
Related to https://oat-sa.atlassian.net/browse/TR-3071

 - After an item renders on Safari iOS, the scrollable elements may not be actually scrollable (a Safari bug), so we force it to reconsider adding the scroll. This seems to be the same issue as [this one](https://stackoverflow.com/questions/28288710/cant-scroll-inside-a-div-after-applying-webkit-transform-in-safari), because it occurs only after a form having the focus AND the loading bar to animate :confused: 

Requires: 
 - [x] https://github.com/oat-sa/tao-core-sdk-fe/pull/130

How to test
 - run unit tests (the keyNavigation tests were already failing :unamused: )
 - along with the other PRs or on TAE by running the failing test on iOS

### DIFF
--- a/src/provider/qti.js
+++ b/src/provider/qti.js
@@ -24,6 +24,7 @@ import $ from 'jquery';
 import _ from 'lodash';
 import __ from 'i18n';
 import cachedStore from 'core/cachedStore';
+import browser from 'util/browser';
 import areaBrokerFactory from 'taoTests/runner/areaBroker';
 import proxyFactory from 'taoTests/runner/proxy';
 import probeOverseerFactory from 'taoTests/runner/probeOverseer';
@@ -38,7 +39,6 @@ import getAssetManager from 'taoQtiTest/runner/config/assetManager';
 import layoutTpl from 'taoQtiTest/runner/provider/layout';
 import states from 'taoQtiTest/runner/config/states';
 import stopwatchFactory from 'taoQtiTest/runner/provider/stopwatch';
-import currentItem from "../helpers/currentItem";
 
 /**
  * A Test runner provider to be registered against the runner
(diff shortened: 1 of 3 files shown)


src/provider/qti.js
```diff
@@ -742,6 +742,17 @@ var qtiProvider = {
 
                     resolve();
                 })
+                .after('render', function() {
+                    //iOS only fix: sometimes the wrapper is not scrollable because of some bugs in Safari iOS
+                    //we have to force it to reconsider if the scroll needs to apply
+                    if (browser.isIOs()) {
```
yukipastelcat, line 748: iOS spelling there is a bit weird, at first I thought about inputs/outputs🤔

yukipastelcat (approved):
Codewise everything looks good, can confirm, that following solution works (tested with Simulator).

shaveko (changes requested):
I've tested on iPad emulator and real iPhone

I'm facing the following issue, which is not critical, but still degrates user experience.

If you navigate and start scrolling while item is being loaded, the scrolling doesn't happen afterwards. Sometimes you wait for a while and content bacomes scrollable again, 

Please check the video attached


https://user-images.githubusercontent.com/10635482/155693723-bd174208-9166-422b-85a4-03043e24415c.mp4

krampstudio:
@shaveko as agreed we will treat this issue separately, thanks for reporting

shaveko (approved):
The issue I found is floating one and doen't happen every time. Also, check the comment to original issue https://oat-sa.atlassian.net/browse/TR-3071?focusedCommentId=174372

We are considering the issue is on Safari side and cannot be worked around by our fix

PR #473, opened 2022-03-11
Fix/TR-3071/Wrong dependency version

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TR-3071

Set the correct dependency version to bring the module `util/browser`.

In a previous [pull request](https://github.com/oat-sa/tao-test-runner-qti-fe/pull/469), the version was updated to align with a companion[ PR](https://github.com/oat-sa/tao-core-sdk-fe/pull/130). Unfortunately, the version was wrong.

**How to test?**
- install the package: `npm i`
- run the unit tests: `npm test`

**Actual result:**
The test `test/provider/qti/test.html` fails

**Expected result:**
The test `test/provider/qti/test.html` succeeds

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2570,9 +2570,9 @@
             "dev": true
         },
         "@oat-sa/tao-core-sdk": {
-            "version": "1.17.0",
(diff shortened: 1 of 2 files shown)


krampstudio (approved):
The dependency is now fixed!

oatymart (approved):
Seems to fix the issue.

PR #476, opened 2022-03-18
Fix/eslint errors

oatymart:
I'm trying to get this repo to zero ESLint errors and warnings, so PR commits will have a green ✅  more often than a red ❌ .

Most of the fixes were simple; I'll comment any where another set of eyes 👀  would be nice.

Out-of-scope: full ES6 conversions of edited files.

To have some more security you can also checkout the middle commit (https://github.com/oat-sa/tao-test-runner-qti-fe/pull/476/commits/f88fae01996eb0f6bcea4067368209ca42c802d1 - changed src but not tests) and run the old tests. Then run the changed tests in the final commit.

### DIFF
--- a/src/helpers/stats.js
+++ b/src/helpers/stats.js
@@ -46,6 +46,7 @@ export default {
                 stats.answered--;
             } else if ((isItemCurrentlyAnswered || sync) && !item.answered) {
                 stats.answered++;
+                // eslint-disable-next-line no-dupe-else-if
             } else if (sync && !isItemCurrentlyAnswered && item.answered && testPart.isLinear) {
                 stats.answered++;
             }
--- a/src/plugins/content/responsiveness/collapser.js
+++ b/src/plugins/content/responsiveness/collapser.js
@@ -257,13 +257,13 @@ export default pluginFactory({
          * @returns {Array}
          */
         function getSortedCollapsiblesFromDom() {
-            var $elements = getControlsFromDom(),
+            var $controlElements = getControlsFromDom(),
                 _allCollapsibles = [],
                 order = {};
 
             // group items by prefix
             // eg. zoomIn and zoomOut -> zoom
-            $elements.each(function() {
+            $controlElements.each(function() {
                 var ctrl = this.dataset.control,
                     // re makes group `foo` from `foo-bar`, `fooBar` and `foo_bar`
                     // if we do not have a prefix use the control name as key to ensure uniqueness
--- a/src/plugins/tools/lineReader/compoundMask.js
+++ b/src/plugins/tools/lineReader/compoundMask.js
@@ -750,6 +750,7 @@ export default function compoundMaskFactory(options, dimensions, position) {
                     .setSize(size.width - options.resizeHandleSize * 2, size.height + options.resizeHandleSize * 2);
             },
 
+            // eslint-disable-next-line no-unused-vars
             beforeResize: function beforeResize(width, height, fromLeft) {
                 this.config.maxWidth = dimensions.rightWidth + (dimensions.innerWidth - constrains.minWidth);
                 this.config.minWidth = constrains.minWidth;
--- a/src/plugins/tools/magnifier/magnifierPanel.js
+++ b/src/plugins/tools/magnifier/magnifierPanel.js
@@ -606,6 +606,7 @@ function magnifierPanelFactory(config) {
 
     dynamicComponentInstance = dynamicComponent({})
         .on('rendercontent', function($content) {
+            // eslint-disable-next-line consistent-this
             var dynamicComponentContext = this;
             var $element = this.getElement();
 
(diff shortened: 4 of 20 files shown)


src/helpers/stats.js
```diff
@@ -46,6 +46,7 @@ export default {
                 stats.answered--;
             } else if ((isItemCurrentlyAnswered || sync) && !item.answered) {
                 stats.answered++;
+                // eslint-disable-next-line no-dupe-else-if
```
oatymart, line 49: ESLint says the final `else if` block can't ever be reached, but I don't think it takes `sync` into account.

src/plugins/content/responsiveness/collapser.js
```diff
@@ -257,13 +257,13 @@ export default pluginFactory({
          * @returns {Array}
          */
         function getSortedCollapsiblesFromDom() {
-            var $elements = getControlsFromDom(),
+            var $controlElements = getControlsFromDom(),
```
oatymart, line 260: `$elements` used in upper scope.

src/plugins/tools/lineReader/compoundMask.js
```diff
@@ -750,6 +750,7 @@ export default function compoundMaskFactory(options, dimensions, position) {
                     .setSize(size.width - options.resizeHandleSize * 2, size.height + options.resizeHandleSize * 2);
             },
 
+            // eslint-disable-next-line no-unused-vars
```
oatymart, line 753: I had to insert a few of these ignores for unused function params. I think it can be nice to leave params to document what the function does/did; if someone disagrees I can remove them though.

src/plugins/tools/magnifier/magnifierPanel.js
```diff
@@ -606,6 +606,7 @@ function magnifierPanelFactory(config) {
 
     dynamicComponentInstance = dynamicComponent({})
         .on('rendercontent', function($content) {
+            // eslint-disable-next-line consistent-this
```
oatymart, line 609: I didn't see another way around using multiple `this` values.

src/proxy/cache/proxy.js
```diff
@@ -356,18 +356,20 @@ export default _.defaults(
...
+                            actions = data;
                             if (data && data.length) {
                                 return this.send('sync', data);
                             }
                         })
                         .catch(err => {
                             if (this.isConnectivityError(err)) {
                                 this.setOffline('communicator');
-                                _.forEach(data, action => {
+                                _.forEach(actions, action => {
```
oatymart, line 372: The only one that looks like a bug before my fix. I refactored this function to the state it was in before `actions` disappeared.

test/plugins/content/itemScrolling/test.js
```diff
@@ -72,8 +71,8 @@ define([
     ];
 
     QUnit.cases.init(pluginApi).test('plugin API ', function (data, assert) {
-        var runner = runnerFactory(providerName);
-        var plugin = pluginFactory(runner);
+        var testRunner = runnerFactory(providerName);
+        var plugin = pluginFactory(testRunner);
```
oatymart, line 75: This whole file was a bit confusing with `testRunner` and (item)`runner` defined in the top scope. I just did the minimum to make it pass.

test/helpers/messages/test.js
```diff
@@ -22,8 +22,8 @@ define(['lodash', 'taoQtiTest/runner/helpers/messages'], function(_, messagesHel
     'use strict';
 
     var messagesHelperApi = [{ title: 'getExitMessage' }];
-    var testActionMessage = 'You will not be able to access this test once submitted. Click \"OK\" to continue and submit the test.';
-    var actionMessage = 'Click \"OK\" to continue.';
+    var testActionMessage = 'You will not be able to access this test once submitted. Click "OK" to continue and submit the test.';
```
bziondik, line 25: In this case - key will be changed, and translations will be lost. Also could be case that it is used in ManualTranslations and for some customer it will be lost

bziondik: sorry, didn't notice :)

bziondik (approved):
- [ ]  New code is covered by tests (if applicable)
- [x]  Tests are running successfully (old and new ones) on my local machine (if applicable)
- [x]  New code is respecting code style rules
- [x]  New code is respecting best practices
- [x]  New code is not subject to concurrency issues (if applicable)
- [ ]  Feature is working correctly on my local machine (if applicable)
- [x]  Acceptance criteria are respected
- [x]  Pull request title and description are meaningful

olga-kulish (approved):
Checks are passing, good job!
I think it's ok to put eslint-disable instead of changing more code, because otherwise we would need to QA it. When somebody will work on this code, he will notice.

src/helpers/stats.js
krampstudio: it looks like eslint is right. To reach the 3rd condition, `isItemCurrentlyAnswered` should be falsy and `item.answered` should be truthy, but with that you'll always reach the 1st block `if (!isItemCurrentlyAnswered && item.answered) {` 
Anyway, it's safer to deactivate the rule.

krampstudio (approved):
awesome!!!

PR #479, opened 2022-04-06
Make tokens able to catch by translation tools

alroniks:
Issue related to a specific bug [REL-667](https://oat-sa.atlassian.net/browse/REL-667).
But actually, it is the wrong usage of localization functions, which looks like technical debt.

### DIFF
--- a/src/helpers/messages.js
+++ b/src/helpers/messages.js
@@ -75,17 +75,11 @@ function getActionMessage(scope, submitButtonLabel = __('OK')) {
         case 'section':
         case 'testSection':
         case 'part':
-            msg = __('Click "%s" to continue', submitButtonLabel).trim();
-            if (msg) {
-                msg += '.';
-            }
-            return msg;
+            return __('Click "%s" to continue.', submitButtonLabel).trim();
         case 'test':
         case 'testWithoutInaccessibleItems':
-            return `${__(
-                'You will not be able to access this test once submitted. Click "%s" to continue and submit the test.',
-                submitButtonLabel
-            )}`;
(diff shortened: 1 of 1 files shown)


tikhanovichA (approved):

src/helpers/messages.js
```diff
@@ -75,17 +75,11 @@ function getActionMessage(scope, submitButtonLabel = __('OK')) {
         case 'section':
         case 'testSection':
         case 'part':
-            msg = __('Click "%s" to continue', submitButtonLabel).trim();
-            if (msg) {
-                msg += '.';
-            }
-            return msg;
+            return __('Click "%s" to continue.', submitButtonLabel).trim();
```
jsconan, line 78: **nitpick:** This will be considered as a new translation token, and the previous one will be removed. It will introduce a regression. I'd advise fixing the related translations, by adding the missing dot to their tokens. Otherwise, they will be lost.

image

image

jsconan (approved):
The change looks straightforward, and it works. However, I don't see the difference in the gathered translations. Not sure it will change the final behavior, apart from the changed token.

Please consider the breaking change it is introducing to the translations.
At least we should auto-correct the translations we have for the product as the change is about a single `.`. It may have an impact on customer-specific translations though.

**Review Checklist**
- [ ] ~~New code is covered by tests (if applicable)~~
- [ ] ~~Tests are running successfully (old and new ones) on my local machine (if applicable)~~
- [ ] ~~New code is not subject to concurrency issues (if applicable)~~
- [x] Pull request title and description are meaningful (it could be better detailed/explained though)
- [x] Commits are atomic

PR #486, opened 2022-05-06
feat/tr 4060/partially answered items in submission

oatymart:
https://oat-sa.atlassian.net/browse/TR-4060

Required by: https://github.com/oat-sa/nfer-enot-template/pull/55
Required by: 

Get the new value `partiallyAnsweredIsAnswered` from test runner config, use it while calculating current item answeredness during response submission.

### DIFF
--- a/src/provider/qti.js
+++ b/src/provider/qti.js
@@ -251,6 +251,7 @@ var qtiProvider = {
             const currentItem = self.getCurrentItem();
             const options = self.getOptions();
             const skipPausedAssessmentDialog = !!options.skipPausedAssessmentDialog;
+            const { partiallyAnsweredIsAnswered } = options.review;
 
             //catch server errors
             var submitError = function submitError(err) {
@@ -317,7 +318,7 @@ var qtiProvider = {
                         // when the test part is linear, the item is always answered as we cannot come back to it
                         const testPart = self.getCurrentPart();
                         const isLinear = testPart && testPart.isLinear;
-                        currentItem.answered = isLinear || currentItemHelper.isAnswered(self);
+                        currentItem.answered = isLinear || currentItemHelper.isAnswered(self, partiallyAnsweredIsAnswered);
                     }
                     resolve();
                 }


jsconan (commented):
It looks ok, but the option `partiallyAnsweredIsAnswered` is not properly handled in the stats helper

src/helpers/stats.js
```diff
@@ -39,9 +39,11 @@ export default {
         const item     = runner.getCurrentItem();
         const testPart = runner.getCurrentPart();
         const stats = _.clone(mapHelper.getScopeStats(map, context.itemPosition, scope));
+        const options = runner.getOptions();
+        const partiallyAnsweredIsAnswered = (options && options.review && options.review.partiallyAnsweredIsAnswered) || true;
```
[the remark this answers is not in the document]
oatymart: Good catch! Thanks

src/provider/qti.js
```diff
@@ -251,6 +251,7 @@ var qtiProvider = {
             const currentItem = self.getCurrentItem();
             const options = self.getOptions();
             const skipPausedAssessmentDialog = !!options.skipPausedAssessmentDialog;
+            const { partiallyAnsweredIsAnswered } = options.review;
```
olga-kulish, line 254: Is it possible for `options.review` to be undefined? If yes, should check for it.

src/provider/qti.js
```diff
@@ -317,7 +318,7 @@ var qtiProvider = {
                         // when the test part is linear, the item is always answered as we cannot come back to it
                         const testPart = self.getCurrentPart();
                         const isLinear = testPart && testPart.isLinear;
-                        currentItem.answered = isLinear || currentItemHelper.isAnswered(self);
+                        currentItem.answered = isLinear || currentItemHelper.isAnswered(self, partiallyAnsweredIsAnswered);
```
olga-kulish, line 321: How do you think, should it influence `"Allow skipping"` validation? Now it allows me to skip partially answered item. I see that customer doesn't mention it in the ticket , so maybe this is completely fine.

jsconan: > Now it allows me to skip partially answered item

What do you mean by this? When skipping an item, it does not matter if it is answered, partially or completely, or not. The answers are set to the default value and that's it.

If you mean the item can be submitted partially answered, yes this is the expected behavior when the option is set.

olga-kulish: I mean, "_A response to this item is required_" dialog does NOT appear if item is answered partially. 
But now I noticed that it already has its own setting for this: `pluginConfig.allowPartial` https://github.com/oat-sa/tao-test-runner-qti-fe/blob/dc4793d0407db480941c3d6e37da94e6cf7899a2/src/plugins/navigation/allowSkipping.js#L89
So this can be closed.

jsconan (approved):
It looks good to me. There is one comment, the same as Olga, but it is not a blocker since the platform config is unlikely to miss the `review` entry from the test runner config.


**Review Checklist**
- [x] New code is covered by tests (if applicable) -- partially unfortunately, the stats helper was not covered by tests
- [ ] ~~New code is not subject to concurrency issues (if applicable)~~
- [x] Commits are atomic

src/provider/qti.js
olga-kulish: Ok, you are right. It's checked in some places like https://github.com/oat-sa/tao-test-runner-qti-fe/blob/develop/src/plugins/navigation/review/review.js#L164 or `src/helpers/stats.js` in this PR, but it shouldn't happen.

olga-kulish (approved):
Good job!

PR #537, opened 2025-09-19
fix: tao-overflow-y in review launch

olga-kulish:
Related to https://oat-sa.atlassian.net/browse/TR-6663


oatymart (approved):
Code review only + seen it in action. Scrollable block looks good.

